PR #2, opened 2015-07-13
Generate imports in sorted path order

peterstace:
This change ensures that the import order is consistent between multiple runs of mockgen (rather than in random map iteration order).  This is useful to help avoid noisy diffs for the case where multiple mocks are regenerated automatically but not all of the mocked interfaces have changed.

### DIFF
--- a/mockgen/mockgen.go
+++ b/mockgen/mockgen.go
@@ -26,6 +26,7 @@ import (
 	"log"
 	"os"
 	"path"
+	"sort"
 	"strconv"
 	"strings"
 	"unicode"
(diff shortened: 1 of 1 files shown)


dsymonds:
Doesn't the generated code get pushed through the `go/format` package, and thus should already be sorted?

peterstace:
The generated code doesn't get pushed through `go/format`, although that way of doing it would be  superior to what I've done here.

I'll close this pull request and put up a new one that uses `go/format`, since the two approaches are significantly different.

PR #5, opened 2015-08-03
Lambda matcher and more validation methods

liron-l:
Lambda matcher allows comparing complex expressions compare instead of
eq or any
-- AtMost and AtLeast call validator (take a look at the following
variety  in the moq library)
http://www.nudoq.org/#!/Packages/Moq/Moq/Times

Signed-off-by: Liron Levin <liron@twistlock.com>

### DIFF
--- a/gomock/call.go
+++ b/gomock/call.go
@@ -20,6 +20,8 @@ import (
 	"strings"
 )
 
+const infinity = 1e8 // close enough to infinity
+
 // Call represents an expected call to a mock.
 type Call struct {
 	t TestReporter // for triggering test failures on invalid call setup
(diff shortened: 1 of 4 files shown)


liron-l:
I signed it!

dsymonds:
Thanks, but we don't need every possible matcher implementation to be in this repo. They can be elsewhere.

liron-l:
Thanks @dsymonds, what do you mean by elsewhere, as a consumer, I would expect a mocking framework to include as much API and utilities as possible. 
Also, comparing using lambda (as opposed to property equality, All or nil) is much more common in practice (at least from my experience).
See for example,
It.Is -Moq (very popular framework for c#) - https://github.com/Moq/moq4/blob/8bc9b111640203af3995fbac98160a5d1f9e1ae2/Source/It.cs
Is.Matching ->
http://ayende.com/wiki/Rhino+Mocks+Constraints.ashx
We can replace lambda with Match or a less confusing term if you think it's more appropriate.

dsymonds:
Gomock does not aspire to be all-inclusive of every possible matcher and utility. They work just as well if they live elsewhere.

PR #9, opened 2015-09-08
add import_prefix to specify a vendored gomock pkg

jmhodges:
The import_prefix flag allows a user of mockgen to specify that they want the gomock library they have vendored (by say, Godeps) to be used instead of the normal one.

It avoids having to remember to call things like `godep save -r` when adding new mocks.

Idea taken from import_prefix in the protoc-gen-go code.

### DIFF
--- a/mockgen/mockgen.go
+++ b/mockgen/mockgen.go
@@ -40,10 +40,11 @@ const (
 )
 
 var (
-	source      = flag.String("source", "", "(source mode) Input Go source file; enables source mode.")
-	destination = flag.String("destination", "", "Output file; defaults to stdout.")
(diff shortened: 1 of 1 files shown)


dsymonds:
golang/go#12278 is probably going to be the right solution. Every tool that operates on Go packages shouldn't hack in their own mechanism for handling vendored packages.

jmhodges:
But that's not guaranteed to be released a stable in 1.6, and folks will still have to support older Go versions for quite a while.

dsymonds:
The plan for the interim will likely to be to implement the logic in a subrepo (e.g. x/tools), which will permit the dozens of relevant tools to use it.

jmhodges:
Just an FYI, we're trying to use gomock in letsencrypt/boulder and are running into issues because of this. We're not able to support vendor until its stable.

What can I do to get this going?

PR #14, opened 2015-12-08
Add mock_prefix flag to specify custom mocked function names

mindaugasrukas:
I have two interfaces in different packages with the same name:

```go
package package1

type Client interface {
    Method1()
}
```

```go
package package2

type Client interface {
    Method2()
}
```

I want to generate mocks for them. The problem is that it generates mocks with the same function names:
```go
func NewMockClient(ctrl *gomock.Controller) *MockClient {
	mock := &MockClient{ctrl: ctrl}
	mock.recorder = &_MockClientRecorder{mock}
	return mock
}
```

Added `mock_prefix` flag allows a user to specify a custom prefix for mocked names:
```go
func NewMockPackage1Client(ctrl *gomock.Controller) *MockPackage1Client {
	mock := &MockPackage1Client{ctrl: ctrl}
	mock.recorder = &_MockPackage1ClientRecorder{mock}
	return mock
}
```

```go
func NewMockPackage2Client(ctrl *gomock.Controller) *MockPackage2Client {
	mock := &MockPackage2Client{ctrl: ctrl}
	mock.recorder = &_MockPackage2ClientRecorder{mock}
	return mock
}
```

### DIFF
--- a/mockgen/mockgen.go
+++ b/mockgen/mockgen.go
@@ -44,6 +44,7 @@ var (
 	destination = flag.String("destination", "", "Output file; defaults to stdout.")
 	packageOut  = flag.String("package", "", "Package of the generated code; defaults to the package of the input with a 'mock_' prefix.")
 	selfPackage = flag.String("self_package", "", "If set, the package this mock will be part of.")
+	mockPrefix  = flag.String("mock_prefix", "", "If set, adds prefix to the mocked function names.")
 
 	debugParser = flag.Bool("debug_parser", false, "Print out parser results only.")
 )
@@ -239,7 +240,7 @@ func (g *generator) Generate(pkg *model.Package, pkgName string) error {
 
 // The name of the mock type to use for the given interface identifier.
 func mockName(typeName string) string {
-	return "Mock" + typeName
+	return "Mock" + *mockPrefix + typeName
 }
 
 func (g *generator) GenerateMockInterface(intf *model.Interface) error {


mindaugasrukas:
I signed it!

dsymonds:
That is not a problem. You aren't meant to combine mocks of two different packages into one package. Keep them in their own packages.

mindaugasrukas:
My example tree structure:
```
--package1
   |-- package1.go
--package2
   |-- package2.go
--package3
   |-- package3.go
   |-- package3_test.go
   |-- package1_mock_test.go
   |-- package2_mock_test.go
```

It makes no sense to keep mocks in their own packages. Also I don't want to include them into final binary - so I named them with `_test.go`. So the problem that `_test.go` files can't be accessed from different package tests.

dsymonds:
It makes perfect sense to put mocks in their own package. That's how gomock is designed to work. You're cutting against the grain.

mindaugasrukas:
So how to access _mock_test.go files from different package?

dsymonds:
Generate the mocks in a regular package (into a package1_mock.go file), and import that package as normal from your test.

mindaugasrukas:
But I want to exclude them from final binary. What is the easiest way of to do that?

mindaugasrukas:
Other solution could be to include some `// +build mock` build constraints..

dsymonds:
I'm not following your confusion. You only need to import the generated mock packages from your test. They won't end up in your final binary.

mindaugasrukas:
Now I understand - your suggestion is to create separate packages just for mocks. I assumed you suggested to move it to real package where interface is. Although I still need to do some workarounds for golint, but it is absolutely fine. Thanks.

PR #17, opened 2015-12-13
narrow types in Recorder method signatures

### LINKED ISSUE #16: calling Recorder methods with numbers causes misses due to Go's untyped constant rules
With an interface

    type Adder interface{
       Inc(x int64) error
     }

the code generated for the MockAdder has the correct int type:

     func (_m *MockAdder) Do(_param0 int64) error {

while the Recorder has `interface{}`s for the argument: 

    func (_m *MockAdderRecoder) Do(arg0 interface{}) error {

This, unfortunately, causes spurious missed expectation in the test case below. This is because the "2" is handed to the AdderRecorder as an `int`. This is Go's untyped int handling code choosing `int` instead of `int64` as its default type since none is provided explicitly.

It looks plausible from the code to generate the `Recorder` methods with the right argument types and convert them down as the call method one does. (But maybe I haven't seen the bug that prevented that in the first place.)

One workaround is to just use the same variable in both places and another is to explicitly convert it to the right type, but it might nice for small tests to just make the right type.

    type Thing struct { a Adder }
    func (t Thing) CallsAdderInc(x int64) { t.a.Inc(x) }

    func TestAdder(t *testing.T) {
    	defer ctrl.Finish()
    	adder := NewMockAdder(ctrl)
    	myThing := Thing{adder}
    	adder.EXPECT().Inc(2).Return(nil) // This 2 is turned into int instead of int64
    	myThing.CallsAdderInc(2)
    }

jmhodges:
This uses the exact types needed in the Recorder methods. It avoids spurious expectation misses caused by Go converting untyped constants to types that do not match the interface generated for.

One example is float literals passed to a `float32` argument were being converted on x86_64 machines to `float64` because the Recoder methods had a `interface{}` arg instead of a `float32` arg.

The code needed already existed in GenerateMockMethod, so we break it out with a small generification around return values.

Fixes #16

### DIFF
--- a/mockgen/mockgen.go
+++ b/mockgen/mockgen.go
@@ -294,13 +294,62 @@ func (g *generator) GenerateMockMethods(mockType string, intf *model.Interface, 
 		g.p("")
 		g.GenerateMockMethod(mockType, m, pkgOverride)
 		g.p("")
-		g.GenerateMockRecorderMethod(mockType, m)
+		g.GenerateMockRecorderMethod(mockType, m, pkgOverride)
 	}
 }
 
 // GenerateMockMethod generates a mock method implementation.
 // If non-empty, pkgOverride is the package in which unqualified types reside.
 func (g *generator) GenerateMockMethod(mockType string, m *model.Method, pkgOverride string) error {
+	rets := make([]string, len(m.Out))
+	for i, p := range m.Out {
+		rets[i] = p.Type.String(g.packageMap, pkgOverride)
+	}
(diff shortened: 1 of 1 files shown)


jmhodges:
I tried working out a test suite for this but it's not quite trivial enough to compile and run test code in process and getting it working with the usual go tools was longer than I could take. It would be cool, though.

dsymonds:
See my comment on #16. It's currently working as intended.

PR #23, opened 2016-01-21
Add extra methods to control the number of times a call is expected.

duskhacker:
The following methods are added:

* MinTimes(int) *Call
* MaxTimes(int) *Call

These additions a user full control over the number of times an
expectation is called.

### DIFF
--- a/gomock/call.go
+++ b/gomock/call.go
@@ -41,8 +41,29 @@ type Call struct {
 	setArgs map[int]reflect.Value
 }
 
+// AnyTimes allows the expectation to be called 0 or more times
 func (c *Call) AnyTimes() *Call {
 	c.minCalls, c.maxCalls = 0, 1e8 // close enough to infinity
+	return c
+}
+
+// MinTimes requires the call to occur at least n times. If AnyTimes or MaxTimes have not been called, MinTimes also
+// sets the maximum number of calls to infinity.
+func (c *Call) MinTimes(n int) *Call {
+	c.minCalls = n
+	if c.maxCalls == 1 {
+		c.maxCalls = 1e8
+	}
+	return c
+}
+
+// MaxTimes limits the number of calls to n times. If AnyTimes or MinTimes have not been called, MaxTimes also
+// sets the minimum number of calls to 0.
+func (c *Call) MaxTimes(n int) *Call {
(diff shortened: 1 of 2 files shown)


dsymonds:
This is too much API. Can we get by with only half of these? The MinTimes/MaxTimes pair should be sufficient, as long as there's a sensible default applied if only one is used.

duskhacker:
Ok, I changed it to be along the lines of what you recommended.

gomock/call.go
```diff
@@ -41,11 +41,32 @@ type Call struct {
 	setArgs map[int]reflect.Value
 }
 
+// AnyTimes allows the expectation to be called 0 or more times
 func (c *Call) AnyTimes() *Call {
 	c.minCalls, c.maxCalls = 0, 1e8 // close enough to infinity
 	return c
 }
 
+// MinTimes sets the minimum number of times an expectation must be called. If maxCalls is at the default of 1
```
dsymonds: Don't refer to unexported details in an exported method's doc comment. A reader of the API isn't going to understand this.

Try something like this:
```go
// MinTimes requires the call to occur at least n times.
// If AnyTimes and MaxTimes have not been called,
// MinTimes also sets the maximum number of calls to infinity.
```

gomock/call.go
```diff
@@ -41,11 +41,32 @@ type Call struct {
...
+// when MinTimes() is called, then maxCalls is set to "infinity". Use MaxTimes() to set the upper bound.
+func (c *Call) MinTimes(n int) *Call {
+	c.minCalls = n
+	if c.maxCalls == 1 {
+		c.maxCalls = 1e8
+	}
+	return c
+}
+
+// MaxTimes sets the maximum number of times an expectation may be called. If minCalls is at the default of 1 when
```
dsymonds: ditto

dsymonds:
Looking fine. Let's just improve the API docs.

dsymonds:
Can you squash the commits when you've addressed my comments too?

duskhacker:
Updated with you latest recommendations, squashed.

dsymonds:
LGTM

dsymonds:
Merged in bd3c8e8

PR #40, opened 2016-06-23
add StrMatcher

noibar:
StrMatcher can be in use when mock expects a complex object, that we want to compare only part of his fields. In order to
do so, you have to implement a string method that represent the
interesting fields of the object. For usage example, look in
matchers_test.go:TestStrMatcher.
TODO: update paths to gomock (instead of my_gomock)


noibar:
I signed it!

noibar:
I signed it!

PR #48, opened 2016-07-29
feature: use "Do" return value. Overlap "Return".

st0nx:
Sometimes need to return a different value for different input arguments. I think it would be more convenient.

Example:

```
func TestRead(t *testing.T) {
    mockCtrl := gomock.NewController(t)
    defer mockCtrl.Finish()
    mockObj := mock_io.NewMockReadWriteCloser(mockCtrl)
    do := func(d []byte) (int, error) {
        return len(d), nil
    }
    mockObj.EXPECT().Read(gomock.Any()).AnyTimes().Do(do)
    fmt.Println(mockObj.Read([]byte{0x1}))
    fmt.Println(mockObj.Read([]byte{0x1, 0x1, 0x1}))
}
```

```
1 <nil>
3 <nil>
```

### DIFF
--- a/gomock/call.go
+++ b/gomock/call.go
@@ -216,7 +216,7 @@ func (c *Call) dropPrereqs() (preReqs []*Call) {
 	return
 }
 
-func (c *Call) call(args []interface{}) (rets []interface{}, action func()) {
+func (c *Call) call(args []interface{}) (action func() []interface {}) {
 	c.numCalls++
 
 	// Actions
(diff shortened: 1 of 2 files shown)


vikstrous:
+1 please take this PR or https://github.com/golang/mock/pull/35 or we'll have to use a private fork or switch away from gomock.

bxy09:
Does there anyone still maintain this tool?
+1 for this PR.

balshetzer:
Thanks for this PR. Something similar has now been pulled in with #126.

PR #50, opened 2016-08-05
Allow generated mocks to pass golint

mb-dev:
The current version of golang/mock generates mocks that fail golint checks. This PR aims to fix that.

For example: I tried generating a mock for a type `GearmanAdmin` and got the following errors:
* comment on exported type MockGearmanAdmin should be of the form "MockGearmanAdmin ..." (with optional leading article)

FIX: I changed the comment to be of that form

* exported function NewMockGearmanAdmin should have comment or be unexported

FIX: I added a comment to that function

* exported method MockGearmanAdmin.EXPECT should have comment or be unexported

FIX: I added a comment on that function

* exported method EXPECT returns unexported type *gearmanadminmock._MockGearmanAdminRecorder, which can be annoying to use

FIX: I made the recorder an exported type, `MockGermanAdminRecorder`. I'll be happy for suggestions on how to fix it in a different way.

* exported method MockGearmanAdmin.PostJob should have comment or be unexported

FIX: Every function in the mock get a comment

### DIFF
--- a/gomock/mock_matcher/mock_matcher.go
+++ b/gomock/mock_matcher/mock_matcher.go
@@ -7,43 +7,49 @@ import (
 	gomock "github.com/golang/mock/gomock"
 )
 
-// Mock of Matcher interface
+// MockMatcher is a Mock of Matcher interface
 type MockMatcher struct {
 	ctrl     *gomock.Controller
-	recorder *_MockMatcherRecorder
+	recorder *MockMatcherRecorder
 }
 
-// Recorder for MockMatcher (not exported)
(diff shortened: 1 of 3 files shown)


mb-dev:
I signed it! (or rather my company - Clever, just signed the CLA)

dsymonds:
Machine-generated code is explicitly out of scope for style (see golint's README). There's much more than just these things that make gomock-generated code poorly styled. This is just churn.

mb-dev:
I can't find it explicitly mentioned in golint README, but this makes sense. I'll exclude *_gen.go from golint.

PR #57, opened 2017-01-14
Add method Invoke  to type Call

satanson:
Although Call.Do can be used to register a doFunc,  however,   doFunc has several flaws:

1. doFunc can has different return types in contrast to the target function.
2. doFunc has no influence on return values of mocked function.
3. closure can be employed to make mocked function return the same values as doFunc, but it's so tricky and nonintuitive.

so another method Invoke is need,  Call.Invoke can be used to register a InvokeFunc which has features as follows:

1. InvokeFunc has the same return types as the target function(receiver type not included).
2. mocked function return the same values  as Invoke.

e.g.

``` go
func TestFoobar(t *testing.T) {
	foobar := mock_gomockdo.NewMockFoobar(ctrl)
		foobar.EXPECT().DoAnyThing(gomock.Any(), gomock.Any()).Invoke(
			func(x, y int) int{
				return x+y
			},
		).AnyTimes()

	for i := 0; i <10; i++ {
		a := foobar.DoAnyThing(1, i)
		fmt.Printf("i=%v, a=%+v\n",i, a)
	}
}
```

### DIFF
--- a/gomock/call.go
+++ b/gomock/call.go
@@ -37,8 +37,12 @@ type Call struct {
 	numCalls int // actual number made
 
 	// Actions
-	doFunc  reflect.Value
-	setArgs map[int]reflect.Value
+	doFunc reflect.Value
+
+	argsType   []reflect.Type
+	retsType   []reflect.Type
+	invokeFunc reflect.Value
+	setArgs    map[int]reflect.Value
 }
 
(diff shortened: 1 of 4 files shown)


satanson:
I signed it!

Issue #52, opened 2016-08-19
gomock fails with unexported interface methods in go 1.7


agtorre:
When working with unexported interface methods, even within the same package, errors are thrown in Go 1.7.

``` go
//bugreport.go
package bugreport

import "fmt"

// Example is an interface with a non exported method
type Example interface {
        someMethod(string) string
}

// CallExample is a simple function that uses the interface
func CallExample(e Example) {
        fmt.Println(e.someMethod("test"))
}
```

``` sh
mockgen -destination  bugreport_mock_test.go -package bugreport -source=bugreport.go Example
```

``` go
//bugreport_test.go
package bugreport

import (
        "testing"

        "github.com/golang/mock/gomock"
)

func TestCallExample(t *testing.T) {
        ctrl := gomock.NewController(t)
        defer ctrl.Finish()

        e := NewMockExample(ctrl)
        e.EXPECT().someMethod(gomock.Any()).Return("it works!")

        CallExample(e)
}
```

``` sh
go test                                                                                                                                                                                                                                                                                                                                                        !24218
--- FAIL: TestCallExample (0.00s)
panic: gomock: failed finding method someMethod on *bugreport.MockExample [recovered]
        panic: gomock: failed finding method someMethod on *bugreport.MockExample [recovered]
        panic: gomock: failed finding method someMethod on *bugreport.MockExample
```

The same test with someMethod written as SomeMethod works fine and as expected. This same test also works in Go 1.6.

lothamerlinden:
It seems like this may be caused by this line in the go 1.7 patch notes:

```The Method and NumMethod methods of Type and Value no longer return or count unexported methods.```

https://github.com/golang/mock/blob/master/gomock/call.go#L260

agtorre:
It seems like this may be caused by this line in the go 1.7 patch notes:

```The Method and NumMethod methods of Type and Value no longer return or count unexported methods.```

https://github.com/golang/mock/blob/master/gomock/call.go#L260

b00lduck:
Any ideas how to solve this? We strongly depend on using 1.7 and gomock with unexported interface methods.

domano:
Bumped into this problem too, is there some way to solve this?

arnaud-lb:
This fixes the problem for me, after re-generating the mocks: https://github.com/golang/mock/issues/59

johnwick03:
@arnaud-lb is this issue fixed? I just got the missing method error for the unexported method.

PR #59, opened 2017-01-16
Fix for #52

arnaud-lb:
This is a quick fix for #52 that works for me, at least. If the general idea is ok for you I'm willing to make it cleaner / make any change that would be requested.


arnaud-lb:
I signed it!

balshetzer:
Thanks for the pull request. I've used the idea in #89 instead.

PR #66, opened 2017-03-13
Make import alias generation predictable

rmohr:
Instead of iterating over a set and generating aliases for conflicting
immports, iterate over a sorted map based on this set. This ensured
predictable output and makes automated checks possible.

We for instance had the problem that our automated check if `go generate` was run, sometimes fails because of different generated code by gomock.

### DIFF
--- a/mockgen/mockgen.go
+++ b/mockgen/mockgen.go
@@ -28,6 +28,7 @@ import (
 	"log"
 	"os"
 	"path"
+	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


rmohr:
I signed it!

fabiand:
@googlebot @rmohr  is covered under the Red Hat CLA.

eparis:
@googlebot he signed it!

rmohr:
I signed it!

PR #67, opened 2017-03-15
Support dynamic return values

dmvk:
Hi,

thanks for an amazing job you guys are doing! It would be really useful if `Call` return values could be changed dynamically as referred here:

https://github.com/golang/mock/issues/34#issuecomment-244477422

Thanks,
D.

### DIFF
--- a/gomock/call.go
+++ b/gomock/call.go
@@ -216,7 +216,7 @@ func (c *Call) dropPrereqs() (preReqs []*Call) {
 	return
 }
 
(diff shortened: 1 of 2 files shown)


dmvk:
I signed it!

sivakku:
Any reason why this fix is not yet merged in? other wise calling Return inside Do() seems broken

balshetzer:
Thanks for this PR. Another solution to this problem has been pulled with #126.

PR #70, opened 2017-04-13
Added usage statement when called with wrong number of arguments (Fixes #69)

### LINKED ISSUE #69: Running mockgen with no arguments does not produce usage statement
There's a very informative usage statement stored in `usage()`, but it's not used when mockgen is invoked with no arguments:

    $ mockgen
    2017/04/13 11:01:21 Expected exactly two arguments

It would be nice if this printed the usage statement too.

TimothyJones:
This is a one line change that fixes #69.

Now when mockgen is invoked with no arguments the usage statement is printed.


TimothyJones:
I signed it!

PR #73, opened 2017-05-31
Add comments for each generated method

rkuris:
Our project gets report card errors because we check in
generated mocks for our end users. This small change
reduces the number of golint errors like this one:

    warning: comment on exported type MockClient
             should be of the form "MockClient ..."
             (with optional leading article) (golint)


rkuris:
I signed it!

balshetzer:
This was also done by #60 which is now merged.

PR #61, opened 2017-02-02
Fix Do method for variadic functions

msabramo:
Without this fix, using the `Do` method on a variadic function will cause a panic, because not all of the arguments will be propagated to the action function:

```
$ go test ./sample
--- FAIL: TestVariadicFunction (0.00s)
panic: reflect: Call using zero Value argument [recovered]
	panic: reflect: Call using zero Value argument [recovered]
	panic: reflect: Call using zero Value argument
...
github.com/golang/mock/gomock.(*Controller).Finish(0xc4200d57a0)
	/gopath/src/github.com/golang/mock/gomock/controller.go:149 +0x320
...
github.com/golang/mock/gomock.(*Call).call.func1()
	/gopath/src/github.com/golang/mock/gomock/call.go:234 +0x6a
...
FAIL	github.com/golang/mock/sample	0.013s
```

Cc: @iderdik, @chr0n1x

PR #74, opened 2017-06-06
Fix call for variable arguments method

hori-ryota:
When executing `Do` with variable arguments, panic because `reflect.Type.NumIn()` of func count variable arguments to `1` .

```
--- FAIL: TestDoWithVariableArguments (0.00s)
panic: reflect: Call using zero Value argument [recovered]
        panic: reflect: Call using zero Value argument
```


balshetzer:
Thanks for the PR. This issue was fixed by #61

PR #75, opened 2017-06-08
add time matcher

klnusbaum:
Times can't be compared using a regular `==`, the `time.Equal` function needs to be used (see https://golang.org/pkg/time/#Time.Equal). I propose we add a special matcher just for `time.Time` values.

### DIFF
--- a/gomock/matchers.go
+++ b/gomock/matchers.go
@@ -17,6 +17,7 @@ package gomock
 import (
 	"fmt"
 	"reflect"
+	"time"
 )
 
 // A Matcher is a representation of a class of values.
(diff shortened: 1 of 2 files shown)


klnusbaum:
I signed it!

balshetzer:
Hi Curtis,

Thanks for the pull request. However, it seems to me that gomock.Eq works fine with time.Time. For example, this program works as expected:

```
package main

import (
        "fmt"
        "time"
)

func main() {
        t := time.Now()
        m := gomock.Eq(t)
        fmt.Println(m.Matches(t))
        fmt.Println(m.Matches(t.Add(time.Second)))
}
```

Let me know if I'm misunderstanding.

Thanks,
Hesky

klnusbaum:
Yea. This was an oversight on my part. My bad. Thanks for taking a look.

PR #78, opened 2017-06-21
Build flags

### LINKED ISSUE #76: Cannot run mockgen with build tags (similar to go build --tags/ginkgo --tags)
We have a project which excludes some C dependencies by running go build as go build --tags "tag1 tag2 tag3". Running ginkgo with ginkgo --tags "tag1 tag2 tag3". However, we couldn't find such option with mockgen. Is there any way to achieve that with mockgen?

balshetzer:
Adds the ability to pass additional flags to `go build`. Fixes #76.

### DIFF
--- a/mockgen/reflect.go
+++ b/mockgen/reflect.go
@@ -34,6 +34,7 @@ import (
 var (
 	progOnly = flag.Bool("prog_only", false, "(reflect mode) Only generate the reflection program; write it to stdout.")
 	execOnly = flag.String("exec_only", "", "(reflect mode) If set, execute this reflection program.")
+	buildFlags = flag.String("build_flags", "", "(reflect mode) Additional flags for go build.")
 )
 
 func Reflect(importPath string, symbols []string) (*model.Package, error) {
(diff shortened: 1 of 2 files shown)


mockgen/reflect.go
```diff
@@ -72,7 +73,7 @@ func Reflect(importPath string, symbols []string) (*model.Package, error) {
 		}
 
 		// Build the program.
-		cmd := exec.Command("go", "build", "-o", progBinary, progSource)
+		cmd := exec.Command("go", "build", *buildFlags, "-o", progBinary, progSource)
```
wsong: I believe that this line is not correct. If buildFlags is empty, you will get an error like:

```
can't load package: package -o: cannot find package "-o" in any of:
/usr/local/go/src/-o (from $GOROOT)
/go/src/-o (from $GOPATH)
```

I think you're missing some quotes here or something.

balshetzer: Yes, sorry about that. This was fixed in #80.

Issue #79, opened 2017-06-22
`go generate` fails after PR #78


korya:
Suddenly my build got broken. `go generate` started to fail with the following error message:

```
can't load package: package -o: cannot find package "-o" in any of:
	/usr/lib/go/src/-o (from $GOROOT)
	/godev/src/-o (from $GOPATH)
can't load package: package prog.bin: cannot find package "prog.bin" in any of:
	/usr/lib/go/src/prog.bin (from $GOROOT)
	/godev/src/prog.bin (from $GOPATH)
can't load package: package prog.go: cannot find package "prog.go" in any of:
	/usr/lib/go/src/prog.go (from $GOROOT)
	/godev/src/prog.go (from $GOPATH)
```

This is the failing go-generate line (it used to work fine):

```go
//go:generate mockgen -package store -destination store-repo-mock_test.go github.com/PlanitarInc/walk-inside-api/service/userstore/repository Userstore
```

korya:
I suspect the problem occurs only when buildFlags argument is not provided. Then `buildFlags` value is the empty string, and hence an additional empty argument is provided to `go build` command.

See https://github.com/golang/mock/pull/78/files#diff-8ba20c3ad6dcdd45830b043e21ba0358R76

PR #80, opened 2017-06-22
Pass --buildFlags value to `go build` only when it is not empty

korya:
The commit fixes https://github.com/golang/mock/issues/79


balshetzer:
Thanks! Sorry for the break.

PR #89, opened 2017-07-21
Fix mocks with unexported methods

### LINKED ISSUE #52: gomock fails with unexported interface methods in go 1.7

balshetzer:
As noted in #52, mocks for interfaces with unexported methods stopped working. A solution was proposed in #59 but I didn't like adding another function to the mock so I'm proposing this alternative, heavily inspired by #59. 

The down side is that it changes the interface to gomock.RecordCall. I don't expect anybody other than the recorder to be calling it but somebody out there probably is doing something fancy and calling it directly. This will break them.

Fixes #52


mockgen/tests/unexported_method/bugreport.go
```diff
@@ -0,0 +1,14 @@
+//go:generate mockgen -destination bugreport_mock.go -package bugreport -source=bugreport.go Example
+package bugreport
```
pasztorpisti: This way `//go:generate` becomes the comment of the package. You can prevent that by inserting an empty line between `//go:generate` and the `package` declaration.

gomock/controller.go
```diff
@@ -80,7 +83,7 @@ func NewController(t TestReporter) *Controller {
 	}
 }
 
-func (ctrl *Controller) RecordCall(receiver interface{}, method string, args ...interface{}) *Call {
+func (ctrl *Controller) RecordCall(receiver interface{}, method string, methodType reflect.Type, args ...interface{}) *Call {
```
pasztorpisti: If backwards compatibility is an issue then it could be maintained by keeping the original method signature and adding a new `RecordCallWithMethodType` (or similar) method which would be used by the generator.

`RecordCall` could simply forward the call to `RecordCallWithMethodType` with a methodType==nil parameter which would force the methodType to be looked up at runtime (like the original implementation).

balshetzer: That's a good suggestion. I'm still leaning towards breaking backwards compatibility here because I think that only the mock and recorder should be calling these functions and the mock won't be using the older version of RecordCall.

pasztorpisti: The breaking change is cleaner and I agree that calling those methods directly doesn't make a lot of sense because the generated mocks are powerful enough.

akshayjshah:
Apparently I'm the somebody doing fancy things :(

I'll open an issue to discuss.

balshetzer:
Sorry about that. I'll restore backwards compatibility.

By the way, are you sure you're doing fancy things or do your mocks just need to be regenerated? Are your mocks generated with `go generate` or some other part of your build process or is it done manually?

gomock/controller.go
balshetzer: @pasztorpisti  Looks like you were right in the first place.

pasztorpisti: I was a bit afraid of this. With a younger or less popular package we'd be in a better position to introduce breaking changes.

PR #90, opened 2017-07-21
Detect adding 'self' as a prereq in Call.After

### LINKED ISSUE #84: Call.After doesn't check for adding a dependency on "self"
The `func (c *Call) After(preReq *Call) *Call` doesn't check/reject buggy invocations when `c == preReq`.

This is a minor problem. I ran into it accidentally by putting individual `Call` instances into variables and manually connecting them using `After()` and `gomock.InOrder()`.

pasztorpisti:
- Adding a `Call` as its own prerequisite will panic (in `Call.After`)
- Written some tests for `Call.After`

Fixes #84

### DIFF
--- a/gomock/call.go
+++ b/gomock/call.go
@@ -156,12 +156,11 @@ func (c *Call) isPreReq(other *Call) bool {
 
 // After declares that the call may only match after preReq has been exhausted.
 func (c *Call) After(preReq *Call) *Call {
+	if c == preReq {
+		c.t.Fatalf("A call isn't allowed to be it's own prerequisite")
+	}
 	if preReq.isPreReq(c) {
-		msg := fmt.Sprintf(
-			"Loop in call order: %v is a prerequisite to %v (possibly indirectly).",
-			c, preReq,
-		)
-		panic(msg)
+		c.t.Fatalf("Loop in call order: %v is a prerequisite to %v (possibly indirectly).", c, preReq)
 	}
 
 	c.preReqs = append(c.preReqs, preReq)
--- a/gomock/call_test.go
+++ b/gomock/call_test.go
@@ -1,0 +1,47 @@
+package gomock
+
+import "testing"
+
+type mockTestReporter struct {
+	errorCalls int
+	fatalCalls int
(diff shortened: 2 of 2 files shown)


balshetzer:
The second commit is interesting. I think we shouldn't work this hard to support all the versions of go. The go team itself doesn't support anything other than the current version and, for major security issues, the previous one. Maybe we should just support the last two releases so that people have some time to migrate.

gomock/call.go
```diff
@@ -156,6 +156,9 @@ func (c *Call) isPreReq(other *Call) bool {
 
 // After declares that the call may only match after preReq has been exhausted.
 func (c *Call) After(preReq *Call) *Call {
+	if c == preReq {
+		panic("a call isn't allowed to be it's own prerequisite")
```
balshetzer: I think it would be better to use c.t.Fatal rather than panic.

pasztorpisti:
I removed the second commit and left only go 1.7 and 1.8 in the travis-ci config.

gomock/call.go
pasztorpisti: Good idea.

Here I simply followed the existing pattern (using a panic like a few lines below). I'm going to switch to `c.t.Fatal` and adjust the tests.

pasztorpisti: Done.

gomock/call_test.go
```diff
@@ -0,0 +1,47 @@
+package gomock
+
+import "testing"
+
+type mockTestReporter struct {
```
pasztorpisti: I could have generated a mock for the `TestReporter` interface but unfortunately in our case this isn't possible:

Since the generated mocks import to gomock package we have to put the mock to a separate package (this is why we have the `gomock/mock_matcher` package) and the gomock package can't import that package because that would cause a circular dependency. This could still be avoided by putting our tests to the gomock_test package (blackbox testing - this is how matchers_test.go uses the mock_matcher package) but we have to keep the tests in the gomock package (whitebox testing) because we have to access the unexported `Call.t` field.

PR #25, opened 2016-02-08
Show file and line number of call during errors

ooesili:
Show file and line number of call during errors

This commit makes erroneous calls show their origin. Unexpected calls
show where they happened. Missing calls and improper expectations show
where they were set up in the tests. This information is printed after
each error message in square brackets.

Here is an example of the output:

```
--- FAIL: TestNoMatchingCall (0.00s)
    controller.go:119: no matching expected call: *mocktest_test.MockDoer.Do([hello]) [/home/ooesili/golang/src/github.com/ooesili/mocktest/mocktest.go:8]
--- FAIL: TestWrongReturnArity (0.00s)
    call.go:83: wrong number of arguments to Return for *mocktest_test.MockDoer.Do: got 1, want 2 [/home/ooesili/golang/src/github.com/ooesili/mocktest/mocktest_test.go:23]
    controller.go:164: missing call(s) to *mocktest_test.MockDoer.Do(is equal to blah) [/home/ooesili/golang/src/github.com/ooesili/mocktest/mocktest_test.go:23]
    controller.go:171: aborting test due to missing call(s)
--- FAIL: TestNotNillable (0.00s)
    call.go:95: argument 0 to Return for *mocktest_test.MockDoer.Do is nil, but string is not nillable [/home/ooesili/golang/src/github.com/ooesili/mocktest/mocktest_test.go:31]
    controller.go:164: missing call(s) to *mocktest_test.MockDoer.Do(is equal to blah) [/home/ooesili/golang/src/github.com/ooesili/mocktest/mocktest_test.go:31]
    controller.go:171: aborting test due to missing call(s)
--- FAIL: TestWrongReturnType (0.00s)
    call.go:105: wrong type of argument 1 to Return for *mocktest_test.MockDoer.Do: string is not assignable to error [/home/ooesili/golang/src/github.com/ooesili/mocktest/mocktest_test.go:39]
    controller.go:164: missing call(s) to *mocktest_test.MockDoer.Do(is equal to blah) [/home/ooesili/golang/src/github.com/ooesili/mocktest/mocktest_test.go:39]
    controller.go:171: aborting test due to missing call(s)
FAIL
FAIL    github.com/ooesili/mocktest 0.002s
```

[...]

PR #65, opened 2017-03-10
Include stacktrace when no matching expected call

RJPercival:
This helps with tracking down the offending call.

PR #97, opened 2017-08-03
Return verbose errors why expected method call did not match

xmik:
Hi. Thanks for gomock! I just started using it and thought it was not enough verbose when it comes to errors messages.

## The benefits
Before this PR, in my tests, I see an error: `no matching expected call`
```
GOROOT=/usr/local/go
GOPATH=/ide/work
/usr/local/go/bin/go test -c -i -o /tmp/TestHTTP_makeRequestUsingCustomClient_in_http_client_test_gogo 34-mock-http-client-with-gomock
/tmp/TestHTTP_makeRequestUsingCustomClient_in_http_client_test_gogo -test.v -test.run ^TestHTTP_makeRequestUsingCustomClient$
	controller.go:129: no matching expected call: *myhttp32.MockHTTPClientInterface.Do([0xc420074b00])
	controller.go:174: missing call(s) to *myhttp32.MockHTTPClientInterface.Do(is equal to &{GET some-urla HTTP/1.1 1 1 map[] <nil> <nil> 0 [] false  map[] map[] <nil> map[]   <nil> <nil> <nil> <nil>})
	controller.go:181: aborting test due to missing call(s)

Process finished with exit code 1
```
after this PR:
```
GOROOT=/usr/local/go
GOPATH=/ide/work
/usr/local/go/bin/go test -c -i -o /tmp/TestHTTP_makeRequestUsingCustomClient_in_http_client_test_gogo 34-mock-http-client-with-gomock
/tmp/TestHTTP_makeRequestUsingCustomClient_in_http_client_test_gogo -test.v -test.run ^TestHTTP_makeRequestUsingCustomClient$
	controller.go:132: no matching expected call: *myhttp32.MockHTTPClientInterface.Do([0xc42000af00]) /ide/work/src/34-mock-http-client-with-gomock/http_client.go:85
		
		The expected argument of index: 0 of this call: /ide/work/src/34-mock-http-client-with-gomock/http_client_test.go:68 did not match the actual argument.
		Actual argument: is equal to &{GET some-urla HTTP/1.1 1 1 map[] <nil> <nil> 0 [] false  map[] map[] <nil> map[]   <nil> <nil> <nil> <nil>}, expected: &{GET some-url HTTP/1.1 1 1 map[] <nil> <nil> 0 [] false  map[] map[] <nil> map[]   <nil> <nil> <nil> <nil>}
	controller.go:177: missing call(s) to *myhttp32.MockHTTPClientInterface.Do(is equal to &{GET some-urla HTTP/1.1 1 1 map[] <nil> <nil> 0 [] false  map[] map[] <nil> map[]   <nil> <nil> <nil> <nil>}) /ide/work/src/34-mock-http-client-with-gomock/http_client_test.go:68
	controller.go:184: aborting test due to missing call(s)

Process finished with exit code 1
```

## This PR contains
  1. The changes that https://github.com/golang/mock/pull/25 introduces. But it had conflicts in relation to [latest master](https://github.com/golang/mock/commit/13f360950a79f5864a972c786a10a50e44b69541) which I fixed.
  2. Changes to `FindMatch` method. It now also returns an error which explains why there was no call match. I added tests for each error explanation message kind.
  3. I removed the square brackets around call origin (path and line to method invocations), because this way, in [Gogland IDE](https://www.jetbrains.com/go/), they are displayed as clickable links.

## You may not like
   * the wording of error messages
   * the errors messages like :"Actual argument: is equal to 15, expected: 3".  It comes from [here](https://github.com/xmik/mock/blob/verbose-errors/gomock/controller_test.go#L302). It would be prettier if the error message was:  "Actual argument: 15, expected: 3". This is because in the [method returning this message](https://github.com/xmik/mock/blob/verbose-errors/gomock/call.go#L205), the `m` object is of type `Matcher`, and it has no public fields, so we can only use its `String()` method. I don't think it is greatly important.

Please voice any concerns and treat me as unexperienced in Go, as I've just recently started using it.

### DIFF
--- a/gomock/call.go
+++ b/gomock/call.go
@@ -29,6 +29,7 @@ type Call struct {
 	methodType reflect.Type  // the type of the method
 	args       []Matcher     // the args
 	rets       []interface{} // the return values (if any)
+	origin     string        // file and line number of call setup
 
 	preReqs []*Call // prerequisite calls
 
@@ -79,8 +80,8 @@ func (c *Call) Do(f interface{}) *Call {
 func (c *Call) Return(rets ...interface{}) *Call {
 	mt := c.methodType
 	if len(rets) != mt.NumOut() {
-		c.t.Fatalf("wrong number of arguments to Return for %T.%v: got %d, want %d",
-			c.receiver, c.method, len(rets), mt.NumOut())
+		c.t.Fatalf("wrong number of arguments to Return for %T.%v: got %d, want %d [%s]",
+			c.receiver, c.method, len(rets), mt.NumOut(), c.origin)
 	}
 	for i, ret := range rets {
 		if got, want := reflect.TypeOf(ret), mt.Out(i); got == want {
@@ -91,8 +92,8 @@ func (c *Call) Return(rets ...interface{}) *Call {
 			case reflect.Chan, reflect.Func, reflect.Interface, reflect.Map, reflect.Ptr, reflect.Slice:
 				// ok
 			default:
-				c.t.Fatalf("argument %d to Return for %T.%v is nil, but %v is not nillable",
-					i, c.receiver, c.method, want)
+				c.t.Fatalf("argument %d to Return for %T.%v is nil, but %v is not nillable [%s]",
+					i, c.receiver, c.method, want, c.origin)
 			}
 		} else if got.AssignableTo(want) {
 			// Assignable type relation. Make the assignment now so that the generated code
@@ -101,8 +102,8 @@ func (c *Call) Return(rets ...interface{}) *Call {
 			v.Set(reflect.ValueOf(ret))
 			rets[i] = v.Interface()
 		} else {
-			c.t.Fatalf("wrong type of argument %d to Return for %T.%v: %v is not assignable to %v",
-				i, c.receiver, c.method, got, want)
+			c.t.Fatalf("wrong type of argument %d to Return for %T.%v: %v is not assignable to %v [%s]",
+				i, c.receiver, c.method, got, want, c.origin)
 		}
 	}
 
@@ -125,7 +126,8 @@ func (c *Call) SetArg(n int, value interface{}) *Call {
 	// TODO: This will break on variadic methods.
 	// We will need to check those at invocation time.
 	if n < 0 || n >= mt.NumIn() {
-		c.t.Fatalf("SetArg(%d, ...) called for a method with %d args", n, mt.NumIn())
+		c.t.Fatalf("SetArg(%d, ...) called for a method with %d args [%s]",
+			n, mt.NumIn(), c.origin)
 	}
 	// Permit setting argument through an interface.
 	// In the interface case, we don't (nay, can't) check the type here.
@@ -134,12 +136,14 @@ func (c *Call) SetArg(n int, value interface{}) *Call {
 	case reflect.Ptr:
 		dt := at.Elem()
 		if vt := reflect.TypeOf(value); !vt.AssignableTo(dt) {
-			c.t.Fatalf("SetArg(%d, ...) argument is a %v, not assignable to %v", n, vt, dt)
+			c.t.Fatalf("SetArg(%d, ...) argument is a %v, not assignable to %v [%s]",
+				n, vt, dt, c.origin)
 		}
 	case reflect.Interface:
 		// nothing to do
 	default:
-		c.t.Fatalf("SetArg(%d, ...) referring to argument of non-pointer non-interface type %v", n, at)
+		c.t.Fatalf("SetArg(%d, ...) referring to argument of non-pointer non-interface type %v [%s]",
+			n, at, c.origin)
 	}
 	c.setArgs[n] = reflect.ValueOf(value)
 	return c
@@ -184,7 +188,7 @@ func (c *Call) String() string {
 		args[i] = arg.String()
 	}
 	arguments := strings.Join(args, ", ")
-	return fmt.Sprintf("%T.%v(%s)", c.receiver, c.method, arguments)
+	return fmt.Sprintf("%T.%v(%s) [%s]", c.receiver, c.method, arguments, c.origin)
 }
 
 // Tests if the given call matches the expected call.
--- a/gomock/call.go
+++ b/gomock/call.go
@@ -163,7 +163,7 @@ func (c *Call) isPreReq(other *Call) bool {
 // After declares that the call may only match after preReq has been exhausted.
 func (c *Call) After(preReq *Call) *Call {
 	if c == preReq {
-		c.t.Fatalf("A call isn't allowed to be it's own prerequisite")
+		c.t.Fatalf("A call isn't allowed to be its own prerequisite")
 	}
 	if preReq.isPreReq(c) {
 		c.t.Fatalf("Loop in call order: %v is a prerequisite to %v (possibly indirectly).", c, preReq)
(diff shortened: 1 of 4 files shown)


xmik:
I signed it!

xmik:
This PR is also related: https://github.com/golang/mock/pull/65 and it could be incorporated instead of or in addition to https://github.com/golang/mock/pull/25.

gomock/controller_test.go
```diff
@@ -187,11 +239,68 @@ func TestUnexpectedArgCount(t *testing.T) {
...
 		// ... so is this.
 		ctrl.Call(subject, "FooMethod")
+	}, "no matching expected call", "Invalid number of arguments of call", "Set: 0, while this call takes: 1")
+	reporter.assertFatal(func() {
+		// The expected call wasn't made.
+		ctrl.Finish()
 	})
+}
+
+func TestExpectedMethodCall_CustomStruct(t *testing.T) {
```
balshetzer: Could you please explain in a comment why we need to test this case?

gomock/call.go
```diff
@@ -192,27 +193,30 @@ func (c *Call) String() string {
 }
 
 // Tests if the given call matches the expected call.
-func (c *Call) matches(args []interface{}) bool {
+// If yes, returns nil. If no, returns error with message explaining why it does not match.
+func (c *Call) matches(args []interface{}) error {
 	if len(args) != len(c.args) {
-		return false
+		return fmt.Errorf("Invalid number of arguments of call: %s. Set: %s, while this call takes: %s",
```
balshetzer: All the error messages returned are worded as error messages to their respective functions. This makes sense in context but I think it doesn't result in the best final error message to the user. Perhaps, instead, they could be worded for the final use. For example. The errors could say something like "Expected call &lt;call details&gt; &lt;mismatch&gt;". Then, the final message will read something like:
Unexpected call &lt;call details&gt; because:
Expected call &lt;call details&gt; has the wrong number of arguments (&lt;got&gt; vs. &lt;want&gt;)
Expected call &lt;call details&gt; doesn't match argument #&lt;index&gt; (&lt;got&gt; vs &lt;want&gt;)

Let me know what you think.

gomock/callset.go
```diff
@@ -47,30 +52,34 @@ func (cs callSet) Remove(call *Call) {
...
 
 	// Search through the unordered set of calls expected on a method on a
 	// receiver.
+	callsErrors := ""
 	for _, call := range calls {
 		// A call should not normally still be here if exhausted,
 		// but it can happen if, for instance, .Times(0) was used.
 		// Pretend the call doesn't match.
 		if call.exhausted() {
 			continue
```
xmik: Added.

gomock/controller.go
```diff
@@ -126,10 +126,10 @@ func (ctrl *Controller) Call(receiver interface{}, method string, args ...interf
 	ctrl.mu.Lock()
 	defer ctrl.mu.Unlock()
 
-	expected := ctrl.expectedCalls.FindMatch(receiver, method, args)
-	if expected == nil {
+	expected, err := ctrl.expectedCalls.FindMatch(receiver, method, args)
+	if err != nil {
 		origin := callerInfo(2)
-		ctrl.t.Fatalf("no matching expected call: %T.%v(%v) [%s]", receiver, method, args, origin)
+		ctrl.t.Fatalf("no matching expected call: %T.%v(%v) [%s]\n%s", receiver, method, args, origin, err)
```
xmik: I agree that it is confusing and I prefer your second suggestion. Continuation in the last comment.

gomock/controller_test.go
```diff
@@ -66,6 +67,18 @@ func (e *ErrorReporter) assertFatal(fn func()) {
...
 			// This is okay - the panic is from Fatalf().
+			if expectedErrMsgs != nil {
+				// assert that the actual error message
+				// contains expectedErrMsgs
+
+				// check the last actualErrMsg, because the previous messages come from previous errors
+				actualErrMsg := e.log[len(e.log)-1]
+				for _, expectedErrMsg := range expectedErrMsgs {
+					if !strings.Contains(actualErrMsg, expectedErrMsg) {
+						e.t.Errorf("Expected the actual error message:\n'%s'\nto contain expected error message:\n'%s'\n", actualErrMsg, expectedErrMsg)
```
xmik: Yes, sure, changed to:
```go
e.t.Errorf("Error message:\ngot: %q\nwant to contain: %q\n", actualErrMsg, expectedErrMsg)
```
So that an example failing test would result in output:
```
controller_test.go:78: Error message:
	got: "Unexpected call to *gomock_test.Subject.FooMethod([argument extra_argument]) at /ide/work/src/github.com/golang/mock/gomock/controller_test.go:89 because \nExpected call at /ide/work/src/github.com/golang/mock/gomock/controller_test.go:238 has the wrong number of arguments. Got: 2, want: 1"
	want to contain: "Invalid number of arguments of call"
```

gomock/controller_test.go
```diff
@@ -119,6 +132,21 @@ func (s *Subject) BarMethod(arg string) int {
...
+type TestStruct struct {
+	Number  int
+	Message string
+}
+
+func (s *Subject) ActOnTestStructMethod(arg TestStruct, arg1 int) int {
+	return 0
+}
+
+// Without this method the string representation of a TestStruct object would be e.g. {%!!(MISSING)s(int=123) no message}
```
xmik: You are right. That comment and function were redundant.

gomock/controller_test.go
xmik: Added 2 comments. If you find this test not needed, I may as well remove it.
   * comment to a similar, but easier test: https://github.com/golang/mock/pull/97/files#diff-763a69b80e2e0fb14b7d265a0f71d31eR190
  * comment to this test: https://github.com/golang/mock/pull/97/files#diff-763a69b80e2e0fb14b7d265a0f71d31eR249

gomock/call.go
xmik: I like your suggestions very much. I implemented them almost as you said. Only in the case of:
```
Expected call <call details> doesn't match argument #<index> (<got> vs <want>)
```
I took the liberty to put `got` and `want` in separate lines, because it seems easier to read when the mocked functions take complex arguments. Example from tests of my project which uses Gomock, where an argument is of type: `http.Request`:
```
controller.go:132: Unexpected call to *myhttp32.MockHTTPClientInterface.Do([0xc42000af00]) at /ide/work/src/34-mock-http-client-with-gomock/http_client.go:85 because: 
	Expected call at /ide/work/src/34-mock-http-client-with-gomock/http_client_test.go:74 doesn't match the argument at index 0.
	Got: &{GET some-url HTTP/1.1 1 1 map[] <nil> <nil> 0 [] false  map[] map[] <nil> map[]   <nil> <nil> <nil> <nil>}
	Want: is equal to &{GET some-urla HTTP/1.1 1 1 map[] <nil> <nil> 0 [] false  map[] map[] <nil> map[]   <nil> <nil> <nil> <nil>}
controller.go:177: missing call(s) to *myhttp32.MockHTTPClientInterface.Do(is equal to &{GET some-urla HTTP/1.1 1 1 map[] <nil> <nil> 0 [] false  map[] map[] <nil> map[]   <nil> <nil> <nil> <nil>}) /ide/work/src/34-mock-http-client-with-gomock/http_client_test.go:74
controller.go:184: aborting test due to missing call(s)
```

I also changed the error message:
```
No expected method calls for that receiver
```
to
```
there are no expected method calls for that receiver
```
so that it reads better with the `Unexpected call <call details> because:`

xmik:
@balshetzer, thank you for all the remarks. They are very helpful. 

I have now one problem: [this line](https://github.com/xmik/mock/blob/verbose-errors/gomock/controller.go#L132) when printed in Gogland, does not result in output containing a clickable link to a particular path and line (the `origin` argument). While [this line](https://github.com/xmik/mock/blob/verbose-errors/gomock/call.go#L204) does. I will try to solve it. (My last commit did not have any impact on it).

xmik:
So the path and line are not a clickable link if they are in the same line as the path and line to `controller.go`. 

E.g. here `/ide/work/src/34-mock-http-client-with-gomock/http_client.go:85` is a link:
```
controller.go:132: Unexpected call to *myhttp32.MockHTTPClientInterface.Do([0xc42000af00]) at 
		/ide/work/src/34-mock-http-client-with-gomock/http_client.go:85 because:
```
whereas here it is not:
```
controller.go:132: Unexpected call to *myhttp32.MockHTTPClientInterface.Do([0xc42000af00]) at /ide/work/src/34-mock-http-client-with-gomock/http_client.go:85 because: 
```

I don't want to dig into this, it is not that crucial.

gomock/controller_test.go
```diff
@@ -20,6 +20,7 @@ import (
 	"testing"
 
 	"github.com/golang/mock/gomock"
+	"strings"
```
pasztorpisti: Cosmetics: std library imports usually reside in the first "import group"

xmik: Ok, @pasztorpisti, I will stick to this advice in the future.

matttproud:
Hi,

I believe this pull request accidentally made the error reporting incur quadratic costs due to the string concatenation in a tight loop routine.  I noticed this when running HEAD of GoMock against a large test.  According to [PProf](https://golang.org/pkg/runtime/pprof/), the problem was this concatenation: https://github.com/golang/mock/blame/4187d4d04aa043124750c9250259ceafdc5f7380/gomock/callset.go#L68-L78

I have a fix out for review here: https://github.com/golang/mock/pull/103

Issue #27, opened 2016-03-09
SetArg on slice silently fails


kaedys:
When SetArg is used to set a slice argument, it silently fails to set the argument when the call occurs.  This is because setting a slice by reflection cannot be done with a one-shot Set() call.  

The diff below adds in support for setting of slices.  A similar method will probably be needed for setting arrays, maps, and possible other types (maybe structs?), though my project doesn't have a need for these, so I didn't investigate their implementation.

``` diff
--- github.com/golang/mock/gomock/call.go   2016-03-09 12:34:55.000000000 -0600
+++ github.com/golang/mock/gomock/call.go   2016-03-09 12:34:56.000000000 -0600
@@ -39,6 +39,13 @@
@@ -212,9 +233,38 @@
        }
        action = func() { c.doFunc.Call(doArgs) }
    }

    for n, v := range c.setArgs {
-       reflect.ValueOf(args[n]).Elem().Set(v)
-   }
+       switch reflect.TypeOf(args[n]).Kind() {
+       case reflect.Slice:
+           setSlice(args[n], v)
+       default:
+           reflect.ValueOf(args[n]).Elem().Set(v)
+       }
+   }

    rets = c.rets
    if rets == nil {
@@ -229,6 +279,15 @@
    return
 }

+func setSlice(arg interface{}, v reflect.Value) {
+   arg = reflect.MakeSlice(v.Type(), v.Len(), v.Cap())
+   av := reflect.ValueOf(arg)
+   for i := 0; i < av.Len(); i++ {
+       av.Index(i).Set(v.Index(i))
+   }
+}
+
 func (c *Call) methodType() reflect.Type {
    recv := reflect.ValueOf(c.receiver)
    for i := 0; i < recv.Type().NumMethod(); i++ {
```

pasztorpisti:
`SetArg` is the sibling of `Return` and meant to be used only with pointer arguments ("out" args) that have a similar purpose as func return values.

While "pointer like objects" (slices, maps) seem to be similar to a normal pointers they can't serve as return values in such a clean way as a pointer arg or a normal func return value - you can't replace the underlying array pointed by the slice at the call site, can't replace the map object referenced by the call site, etc... - you can only modify them. For this reason if someone wants to return a slice or a map then the arg should be a slice pointer (e.g.: `*[]int`) or map pointer - or better - just a plain func return value.

I think the correct way to fix this bug is explicitly checking whether the given arg is a pointer (instead of assuming) and failing if it isn't.

[...]

PR #98, opened 2017-08-14
Support slices in SetArg()

hatstand:
I needed this while trying to write tests against https://godoc.org/github.com/kidoman/embd#SPIBus

`SPIBus::TransferAndReceiveData([]byte)` writes its output back to the input byte slice.

Would also fix https://github.com/golang/mock/issues/27

### DIFF
--- a/gomock/call.go
+++ b/gomock/call.go
@@ -142,8 +142,10 @@ func (c *Call) SetArg(n int, value interface{}) *Call {
 		}
 	case reflect.Interface:
 		// nothing to do
+	case reflect.Slice:
+		// nothing to do
 	default:
-		c.t.Fatalf("SetArg(%d, ...) referring to argument of non-pointer non-interface type %v [%s]",
+		c.t.Fatalf("SetArg(%d, ...) referring to argument of non-pointer non-interface non-slice type %v",
 			n, at, c.origin)
 	}
 	c.setArgs[n] = reflect.ValueOf(value)
@@ -243,7 +245,12 @@ func (c *Call) call(args []interface{}) (rets []interface{}, action func()) {
 		action = func() { c.doFunc.Call(doArgs) }
 	}
 	for n, v := range c.setArgs {
-		reflect.ValueOf(args[n]).Elem().Set(v)
+		switch reflect.TypeOf(args[n]).Kind() {
+		case reflect.Slice:
+			setSlice(args[n], v)
+		default:
+			reflect.ValueOf(args[n]).Elem().Set(v)
+		}
 	}
 
 	rets = c.rets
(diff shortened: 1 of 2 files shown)


hatstand:
I signed it!

balshetzer:
I'm not sure about this. This overloads SetArg to also fill a slice, which feels like it's not the same thing.

This need shouldn't block anyone because the easy workaround is to use the Do function to fill the slice. However, I could see the value of a function specifically for filling an input slice.

hatstand:
I think it's morally the same as other uses of SetArg as it does actually
change the input argument.

Happy to implement it as a separate function though; name suggestions?

pasztorpisti:
I agree with @balshetzer that it's better to handle only pointer args in `SetArg`.

Pointer args are sometimes used as "return values" - probably by ex C programmers - despite the fact that it isn't very idiomatic in go where you can easily return multiple values.

Any other case - like modifying an array or any other input object - isn't really like simply returning a value. Modifying an incoming object is a messy business that is better done in `Do`. In many cases you can't even perform the full spectrum of operations on the incoming object when you want to modify it - e.g.: you can add extra items to the slice because that might involve allocation and the caller would still have a reference only to the original underlying array.

The above problems have been discussed in #27.

In my opinion `SetReturnArg` or `SetOutArg` could be a better name for this function for the above reasons (but renaming isn't an option because of backward compatibility). We could perhaps add a godoc comment to the function better explaining its purpose.

pasztorpisti:
I agree that `Do` is a bit inflexible though and I was thinking about this problem a few weeks ago. You have to write the full function signature and it doesn't support return values.

One of my ideas was extending the current behaviour of `Do` and being able to pass a generic handler function to it that always has the same signature, something like `func(args map[string]interface{})`.

If `Do` supported this generic interface then it would be quite easy to write reusable project-local test utility functions to deal with issues, e.g.:

```go
func NewArrayFirstByteZeroer(argName string) func(args map[string]interface{}) {
	return func(args map[string]interface{}) {
		args[argName].([]byte)[0] = 0
	}
}
```

... and then reusing it would be a breeze:

```go
myMock.EXPECT().Whatever().Do(NewArrayFirstByteZeroer("data"))
```

Since the arg names are available only in case of source generation method (and not in case of reflection based) it might make sense to use the `func(args ...interface{})` signature instead of `func(args map[string]interface{})`.

hatstand:
I guess the more accurate name for `SetArg` would have been `SetArgumentPointee` like gmock. gmock also has `SetArrayArgument` if you wanted to follow their lead.

pasztorpisti:
You can't set an array arg, you can only modify it.

I wouldn't put any new "features" into gomock other than improving `Do` to provide a way for people to handle edge cases. Most mocking libraries are a mess full of narrow purpose functions.

balshetzer:
@hatstand I've thought about it and would like to accept this PR. Would you be willing to add unit tests and update the comment on SetArg to explain what it does?

@pasztorpisti I hear your points too but populating arg slices is clearly something that is done in Go and so, I think, having SetArg "do the right thing" for slices seems like it will make it more useful without much cost.

pasztorpisti:
Populating arg slices usually involves complex logic (which is basically simulating the mocked logic - something for which `Do` was invented). Slice population isn't as simple as a `Return()`.

hatstand:
Added documentation & unit tests.

PR #99, opened 2017-08-15
Make mockgen mock dependencies in vendor dir

### LINKED ISSUE #4: mocking dependencies located in a vendor directory

anfernee:
Fix #4

### DIFF
--- a/mockgen/model/model.go
+++ b/mockgen/model/model.go
@@ -23,6 +23,9 @@ import (
 	"strings"
 )
 
+// pkgPath is the importable path for package model
+const pkgPath = "github.com/golang/mock/mockgen/model"
+
 // Package is a Go package. It may be a subset.
 type Package struct {
 	Name       string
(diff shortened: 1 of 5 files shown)


balshetzer:
This is similar to #28 but has the benefit of not having any conflicts and adds a test. #28 also addresses types in the vendored package. Does this PR do that too? There are also a couple of other differences but I'm not sure what the impact of those are.

The reason I haven't, yet, approved either is that I need to take some time to fully understand the way they solve the problem and the possible impact of the changes on existing users. 

For example, the issue of the code being in read-only directories was raised in #28. I'm not sure if this is hypothetical or a practical concern.

anfernee:
Thanks @balshetzer. yes, this solves the same issue. The difference is it didn't hack the `PkgPath` result to remove the `vendor` part, it tried to "correct" its behavior but registering the gob with the right name. 
I am not sure about a read-only tree. Eventually mockgen is going to write the mock files in the project most of the time, so it should be writable. Again, I am not sure.

anfernee:
@balshetzer I almost forgot this PR. Any concerns?

PR #100, opened 2017-08-16
mockgen: emit package docs comments if needed.

matttproud:
This commit enhances the cosmetic output of the mockgen code generator
such that its output includes a package-level comment when the generator
writes to a brand-new package.  This prevents clobbering the package
documentation from pre-existing packages.

This change is motivated by several things:

(1.) enhancing the readability of http://godoc documentation for large
     package trees.

(2.) providing a differentiable characteristic from other
     human-written packages.  This helps both humans and programs
     differentiate between packages themselves for purposes of
     classification, which is useful when automatically refactorings
     across an entire codebase like a monorepo.

(3.) keeping the output stylistically consistent with that emitted from
     the protoc-gen-go code generator, which Protocol Buffers uses.

I have already signed the Google and Go CLAs and am an employee at @google.  I
plan on using this internally within our third_party hierarchy and have two
immediate use cases today where I need to update our documentation corpus-wide.


mockgen/mockgen.go
```diff
@@ -225,6 +225,12 @@ func (g *generator) Generate(pkg *model.Package, pkgName string) error {
 		localNames[pkgName] = true
 	}
 
+	// Write a package-level documentation comment only if emitting a new
+	// package outright, which would not clobber the original's
+	// documentation.
+	if inPkg, outPkg := pkg.Name, pkgName; inPkg != outPkg {
```
balshetzer: This test is not sufficient to detect if we're outputting a new package or not. How concerned do you think we should be about clobbering existing package documentation?

matttproud: Hi, @balshetzer, thank you for the immediate attention!

I imagined it wouldn't be 100 percent comprehensive.  I would be gravely concerned from a developer experience perspective if GoMock overwrote *without trying* to check whether it would overwrite.

Looking at our monorepo, I estimate (by informal sampling) …

1. we explicitly set an output package name 50 percent of time

2. we overwhelmingly set the output package name to a divergent name from the original with the nomenclature for these being all across the board — ca. 95 percent of the time

Are there any cheap ways we could further enhance this without wandering around the filesystem aimlessly?  We use [Bazel](https://bazel.build) extensively, so a free filesystem walk is out of the question, since we must accurately declare build inputs and their dependencies in advance.

Alternatively, this check could be sufficient given pt. 2.  From a perspective of good Go design, I cannot imagine too many folks bundling GoMock output with their production code's exported types in the same package.  The reason is cluttering the minimum viable set of types required to use a package.  Perhaps if someone attempts to to emit to a pre-existing package of production export typed — considering the best practice I highlighted — then they should be able to accept the risks of our good faith attempt.

IIUC, godoc choses the package's documentation from the first file it visits in the directory which contains a package-level documentation comment.

matttproud: Taking a comparative approach:

The ubiquitous goprotobuf [emits the package documentation arbitrarily without checking](https://github.com/golang/protobuf/blob/157d9c53be5810dd5a0fac4a467f7d5f400042ea/protoc-gen-go/generator/generator.go#L1249) whether such output would collide. 

Like Protocol Buffers, folks generally wouldn't mix GoMock outputs with production, hand-written packages.  I am inclined to say we just emit it always or keep the good faith check — imperfections and all.

Thoughts?

balshetzer: Another option would be to add something like a --write_package_comment flag and just drive the behavior on that. We could even default it to true if that's the more likely desired behavior. Would that be better?

matttproud:
@balshetzer, a flag will work fine.  I have adjusted the pull request as you propose and updated the new test fixtures.  I can — I expect — easily handle the overwrite behavior in [Bazel Macros](https://docs.bazel.build/versions/master/skylark/macros.html).

matttproud:
Thank you for the review and attention!  :)

PR #101, opened 2017-08-18
Fix variadic mock for gomock.Any() . fix #68

### LINKED ISSUE #68: Support variadic methods
I have a method with a printf-like signature: func(string, ...interface{}).  When this is mocked and set up with myMock.EXPECT().myFunc(gomock.Any()), then the function is actually called, Call.call panics because len(args) < ft.NumIn().  This is because the generated mock code builds args by flattening the variadic arguments and normal arguments into a single slice, but the reflection API expects the variadic arguments to be wrapped up in a slice.

hori-ryota:
Please see #68 .


hori-ryota:
Sorry, broke other tests. I'll fix.

hori-ryota:
fixed.

hori-ryota:
@balshetzer how about this?

hori-ryota:
fmm...

balshetzer:
Hi. This change is good but I wanted to clean it up a bit. The first error message is not the same as the others, for example.

hori-ryota:
OK. Is something better for me to do?

hori-ryota:
@balshetzer Can I fix it? Let me know what you want to fix and I will try!

balshetzer:
@hori-ryota The only mandatory thing I want fixed is for the error message on line 203 to be clearer. It should be similar to the other error messages.

Some nice to have improvements:
- the error case mentioned above should probably fail when the EXPECT call is first created.
- there are many error cases here. Can they be reduced?
- The condition on lines 222-223 is complex and not obvious. It might be good to have some comments with examples to explain the various situations it covers.
- alternatively, could the matching be rewritten to first handle all non-variadic args and then have a separate case to handle the variadic args? That might make the code easier to read.

hori-ryota:
How about my refactor?

balshetzer:
Thanks!

PR #107, opened 2017-09-14
retain ordering of expected calls

natebrennand:
While writing tests for a stateful object, I noticed unexpected behavior
where the mock would use the 1st call, but then the 3rd expected call
before the 2nd (called in [1,3,2]).

Since our tests all used the `gomock.Any()` function param, there was no
ordering that could be detected by the matching without chaining all
calls with `After()`.


balshetzer:
@natebrennand It occurred to me after merging this that I should have asked for a unit test for this. I don't have any more leverage but I would appreciate it if you would write one.

Issue #110, opened 2017-09-20
Mocks cannot safely be passed to other goroutines


jimpo:
If a mock expectation fails, it attempts to fail the test using testing.T.Fatalf (see [here](https://github.com/golang/mock/blob/45f67fc6bd1ec393ae6c0145e35d67b3867bbe96/gomock/controller.go#L132)). Fatalf in turn exits the current goroutine with `runtime.Goexit()`. However, if the mock is being using in a goroutine other than the testing thread, it will not exit the testing thread. Even worse, it can cause the current goroutine to exit in an unexpected way which in my application skips cleanup and causes a deadlock in the test. While I could refactor a bunch of stuff to do cleanup with defers, I see no benefit other than accommodating this behavior of gomock at the expense of additional complexity. Are there any known solutions?

balshetzer:
gomock calls Fatalf on the [TestReporter](https://github.com/golang/mock/blob/master/gomock/controller.go#L67) that you pass to the Controller's constructor. That doesn't have to be a testing.T. You can create your own TestReporter that does whatever you want when Fatalf is called.

balshetzer:
Please let me know if these suggestions resolved your issue or not. If yes, let's close this. If not then please explain so we can try to resolve it.

jimpo:
https://github.com/golang/mock/pull/111 provides an acceptable workaround for me, so I'll close.

PR #111, opened 2017-09-20
gomock.WithContext: controller with a context

abbot:
gomock.WithContext: controller with a context which is cancelled on Fatals.

testing.T.FailNow (and Fatal, ...) work only when called from the same
goroutine as the test. This makes using gomock with e.g. RPC server
mocks quite a challenge as the tests are likely to deadlock instead of
failing properly.

This adds a new constructor returning an additional context, which is
cancelled on any fatal failure in mock.

This provides a reasonable workaround for #110 and similar issues.


jimpo:
This looks like a pretty decent solution to me. Thanks for getting a proposed fix out so quickly.

PR #112, opened 2017-09-22
possibility to specify custom mock names

mariusstaicu:


mariusstaicu:
I signed it!

mariusstaicu:
I signed it!

vborcea:
nice work

mockgen/mockgen.go
```diff
@@ -98,13 +99,28 @@ func main() {
...
 	if _, err := dst.Write(g.Output()); err != nil {
 		log.Fatalf("Failed writing to destination: %v", err)
 	}
 }
+func parseMockNames(names string) map[string]string {
+	mocksMap := make(map[string]string)
+	for _, kv := range strings.Split(names, ",") {
+		parts := strings.SplitN(kv, "=", 2)
+		if len(parts) != 2 || parts[1] == "" {
+			log.Fatalf("bad mock names spec: %v", kv)
```
vborcea: If the parts are not OK you should continue with the next part and drop the current "bad" ones

mariusstaicu: I would rather exit with error if that happens.

vborcea: Ok

mariusstaicu:
I signed it!

PR #115, opened 2017-10-04
Dynamic return values

sivakku:
Dynamically return values based on the processing on the inputs. Added unit test as well.
This is same as earlier PR request #67 but with an added unit test.

### DIFF
--- a/gomock/call.go
+++ b/gomock/call.go
@@ -296,7 +296,7 @@ func (c *Call) dropPrereqs() (preReqs []*Call) {
 	return
 }
 
-func (c *Call) call(args []interface{}) (rets []interface{}, action func()) {
+func (c *Call) call(args []interface{}) (action func()) {
 	c.numCalls++
 
 	// Actions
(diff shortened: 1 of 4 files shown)


sivakku:
I signed it!

balshetzer:
Thanks for this PR.

It looks like the continuous build failed because the files need to have go fmt run on them.

That aside, I'm not a huge fan of using this technique for dynamically returning results. I think that the return value of the function passed into Do() should have been used as the return value of the function. However, making that change would break backwards compatibility. 

As a close second, I'd interested in adding a new DoAndReturn (or a better name) that does this.

sivakku:
@balshetzer - Thanks for pointing out about gofmt, from travis CI it wasn't clear to me, couldn't find logs on it.

Coming from using other mock frameworks(even in GO) which uses function to return values, that seems more intutive, but after reading through this thread #34 the gomock designers doesn't seem to prefer that way.

i liked your suggestion about DoAndReturn(), i will send that as a separate PR after this.

sivakku:
@balshetzer - Can this change be merged?

balshetzer:
Hi Sivakumar,

Thanks for your Pull Request. However, I've decided to use a different approach in #126 to fix this problem.

sivakku:
No worries, we are in better place now. :)

PR #118, opened 2017-10-19
fix backwards compatibility with golang < 1.7

jeking3:


jeking3:
I signed it!

jeking3:
For reference, this change fixed the Apache Thrift CI builds which run on Go 1.4.2, 1.6.2, and 1.8.3 in containers.  Without this change, only the 1.8.3 build worked.  With this change, golang.org/x/net/context is must be "go"tten as part of the build process.

PR #120, opened 2017-10-20
add travis ci builds for patches as well as 1.6 to detect backwards compatibility issues

jeking3:


jeking3:
It looks like golang itself doesn't have a mechanism to do a smooth migration of experimental to official library status.  In perl you can do something like "use OfficialLibv2 || use ExperimentalLibv1" and as long as they are API compatible this works.  I added more to the gomock CI build and it fails building on 1.6 because it looks like other libraries use "context" now.  It seems to me that folks want everyone to move off 1.6 ASAP.  While go might be backwards compatible, handling experimental -> official library migration is not.

The way the Apache Thrift project handles this issue is that it detects the go version during the build and then includes either a "go17.go" file or a "pre-go17.go" file and they each have a different import directive and then alias context to something else.

It would be way more powerful if the language could handle a failed import by trying another one.  I think I'll bring this up on the golang project itself.

ianlancetaylor:
If you are using the go tool to build, you can use build tags to select whether to build the Go 1.7 version of the pre-Go 1.7 version.  You don't need a separate mechanism.

jeking3:
@ianlancetaylor Thanks - it may be that way, I haven't looked closely enough.

balshetzer:
I think 1.6 is failing because your last PR missed one of the uses of "context". I've fixed that in #121. Could you try removing the allow_failure and see if it passes now?

jeking3:
I wasn't looking at the sample code, that's why. :)  I rebased, but I didn't remove the allow_failures stanza for now.

balshetzer:
OK. FYI: it's failing now because 1.6 doesn't have t.Run and that's used in some of the tests.

PR #126, opened 2017-11-09
Add DoAndReturn

### LINKED ISSUE #34: return value depend on input value

balshetzer:
Add DoAndReturn as a complement to Do that does not ignore its return values.

Refactored how all the call actions work to make this easier.

Fixes #34

### DIFF
--- a/gomock/call.go
+++ b/gomock/call.go
@@ -25,12 +25,11 @@ import (
 type Call struct {
 	t TestReporter // for triggering test failures on invalid call setup
 
-	receiver   interface{}   // the receiver of the method call
-	method     string        // the name of the method
-	methodType reflect.Type  // the type of the method
-	args       []Matcher     // the args
-	rets       []interface{} // the return values (if any)
-	origin     string        // file and line number of call setup
+	receiver   interface{}  // the receiver of the method call
+	method     string       // the name of the method
+	methodType reflect.Type // the type of the method
+	args       []Matcher    // the args
+	origin     string       // file and line number of call setup
 
 	preReqs []*Call // prerequisite calls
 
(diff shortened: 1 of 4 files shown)


gomock/call.go
```diff
@@ -70,11 +70,56 @@ func (c *Call) MaxTimes(n int) *Call {
 	return c
 }
 
-// Do declares the action to run when the call is matched.
+// DoAndReturn declares the action to run when the call is matched.
+// The return values from this function are returned by the mocked function.
+// It takes an interface{} argument to support n-arity functions.
+func (c *Call) DoAndReturn(f interface{}) *Call {
+	// TODO: Check arity and types here, rather than dying badly elsewhere.
+	v := reflect.ValueOf(f)
```
andreecmonette: Since the code is structured this way now we can verify that v.Kind() != reflect.Func and c.t.Fatalf if not; we can also explicitly enforce that len(args) == v.NumIn() below instead of panicking in reflect. Feel free to keep the latter as a todo/do it in a later commit.

gomock/call.go
```diff
@@ -70,11 +70,56 @@ func (c *Call) MaxTimes(n int) *Call {
...
+}
+
+// Do declares the action to run when the call is matched. The function's
+// return values are ignored to retain backward compatibility. To use the
+// return values call DoAndReturn.
 // It takes an interface{} argument to support n-arity functions.
 func (c *Call) Do(f interface{}) *Call {
 	// TODO: Check arity and types here, rather than dying badly elsewhere.
-	c.doFunc = reflect.ValueOf(f)
+	v := reflect.ValueOf(f)
```
andreecmonette: Same comment as line 78.

josh-tepper:
@balshetzer  -- I apologize for not writing back sooner -- I was pulled away for a high priority work project, and by the time I returned, this was already merged.  Nonetheless, I wanted to share some thoughts:

I particularly like the way this was implemented.  The idea of turning everything into an action is an elegant simplification of the code.  Further, the idea that actions are simply executed in order means that DoAndReturn can combine nicely with Do, Return, etc.  Other calls to Do(..) will still run, and whichever call -- Return or DoAndReturn -- is called later will determine the ret value.
 
+1 for andreecmonette's comments about checking the function's type when DoAndReturn is called.

PR #127, opened 2017-11-10
Fix varargs matching.

### LINKED ISSUE #125: Error when using variadic function signature
I have a service I am mocking like this:
```
type RoleService interface {
	GetScopes(roles ...string) (result []string)
}
```

When my test calls:
```
roleService.EXPECT().
	GetScopes(gomock.Eq([]{"admin"})).
	Return([]string{"a:foo", "b:bar", "admin:admin"}).
	Times(1)
```

I get an error:
```
Unexpected call to *mock_service.MockRoleService.GetScopes([admin]) at bitbucket.org/xxxxxxx/yyyyyyyy/service/_test/_obj_test/auth0_service.go:266 because:
      Expected call at /mnt/e/work/src/bitbucket.org/xxxxxxxxx/yyyyyyy/service/auth0_service_test.go:337 doesn't match the argument at index 0.
      Got: admin
      Want: is equal to [admin]
```

If I switch to using []string in the function signature, it works fine. What am I missing here?

balshetzer:
Two fixes:
- Allow matching varargs by slice even when there is only one vararg.
- Allow matching varargs by slice using Eq. For example Eq([]int{1, 2}).

Fixes #125

### DIFF
--- a/gomock/call.go
+++ b/gomock/call.go
@@ -245,7 +245,7 @@ func (c *Call) matches(args []interface{}) error {
 				// sample: Foo(a int, b int, c ...int)
 
 				if len(c.args) == len(args) {
-					if c.args[i].Matches(args[i]) {
+					if m.Matches(args[i]) {
 						// Got Foo(a, b, c) want Foo(matcherA, matcherB, gomock.Any())
 						// Got Foo(a, b, c) want Foo(matcherA, matcherB, someSliceMatcher)
 						// Got Foo(a, b, c) want Foo(matcherA, matcherB, matcherC)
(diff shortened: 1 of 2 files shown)


gomock/call.go
```diff
@@ -245,15 +245,26 @@ func (c *Call) matches(args []interface{}) error {
...
 						break
 					}
-				} else if c.args[i].Matches(args[i:]) {
+				}
+				// The number of actual args don't match the number of matchers.
+				// If this function still matches it is because the last matcher
+				// matches all the remaining arguments or the lack of any.
+				// Convert the remaining arguments, if any, into a slice of the
+				// expected type.
+				vargsType := c.methodType.In(c.methodType.NumIn() - 1)
```
andreecmonette: I think this should be c.methodType.In(c.methodType.NumIn() - 1).Elem() otherwise 263 is creating a slice of slices.

balshetzer: It turns out that MakeSlice wants a slice type.

Issue #130, opened 2017-11-21
Panic during varargs handling


aaronbee:
After upgrading to the latest gomock HEAD our test hit a panic:

```
panic: reflect.Set: value of type *gen.Routes is not assignable to type grpc.CallOption

goroutine 7 [running]:
reflect.Value.assignTo(0x150e040, 0xc42005f680, 0x16, 0x1563786, 0xb, 0x14d2780, 0xc42017d400, 0x0, 0x1012e40, 0x10)
	/usr/local/Cellar/go/1.9.2/libexec/src/reflect/value.go:2192 +0x3a6
reflect.Value.Set(0x14d2780, 0xc42017d400, 0x194, 0x150e040, 0xc42005f680, 0x16)
	/usr/local/Cellar/go/1.9.2/libexec/src/reflect/value.go:1357 +0xa4
reflect.Append(0x1491760, 0xc420196000, 0x97, 0xc42004f678, 0x1, 0x1, 0x97, 0x10494ac, 0x17ca8d8)
	/usr/local/Cellar/go/1.9.2/libexec/src/reflect/value.go:1840 +0xdd
arista/vendor/github.com/golang/mock/gomock.(*Call).matches(0xc42008f130, 0xc420153fc0, 0x2, 0x2, 0x1, 0xc42018e028)
	/Users/aaronb/go/src/arista/vendor/github.com/golang/mock/gomock/call.go:354 +0xace
arista/vendor/github.com/golang/mock/gomock.callSet.FindMatch(0xc4201581e0, 0xc420158210, 0x15205c0, 0xc420045320, 0x1565ab0, 0x10, 0xc420153fc0, 0x2, 0x2, 0xc42001e000, ...)
	/Users/aaronb/go/src/arista/vendor/github.com/golang/mock/gomock/callset.go:73 +0x1cd
arista/vendor/github.com/golang/mock/gomock.(*Controller).Call.func1(0xc420152520, 0x15205c0, 0xc420045320, 0x1565ab0, 0x10, 0xc420153fc0, 0x2, 0x2, 0x0, 0x0, ...)
	/Users/aaronb/go/src/arista/vendor/github.com/golang/mock/gomock/controller.go:146 +0x10f
arista/vendor/github.com/golang/mock/gomock.(*Controller).Call(0xc420152520, 0x15205c0, 0xc420045320, 0x1565ab0, 0x10, 0xc420153fc0, 0x2, 0x2, 0x0, 0x0, ...)
[...]

PR #131, opened 2017-11-21
Fix panic during varargs matching

aaronbee:
This breaks varargs matching a slice that has only one element. I had to remove the test case for this.

Fixes https://github.com/golang/mock/issues/130

### DIFF
--- a/gomock/call.go
+++ b/gomock/call.go
@@ -328,55 +328,55 @@ func (c *Call) matches(args []interface{}) error {
 		}
 
 		for i, m := range c.args {
-			if i == len(c.args)-1 {
-				// The last arg has a possibility of a variadic argument, so let it branch
-
-				// sample: Foo(a int, b int, c ...int)
-
-				if len(c.args) == len(args) {
-					if m.Matches(args[i]) {
-						// Got Foo(a, b, c) want Foo(matcherA, matcherB, gomock.Any())
-						// Got Foo(a, b, c) want Foo(matcherA, matcherB, someSliceMatcher)
-						// Got Foo(a, b, c) want Foo(matcherA, matcherB, matcherC)
-						// Got Foo(a, b) want Foo(matcherA, matcherB)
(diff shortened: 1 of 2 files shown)


aaronbee:
I'm bugging people to get added to my company CLA.

aaronbee:
My new commit fixes support for matching a slice of one element.

aaronbee:
I signed it!

gomock/controller_test.go
```diff
@@ -132,7 +133,7 @@ func (s *Subject) BarMethod(arg string) int {
 	return 0
 }
 
-func (s *Subject) VariadicMethod(arg ...string) {}
+func (s *Subject) VariadicMethod(arg int, vararg ...string) {}
```
balshetzer: What does this change do?

gomock/call.go
```diff
@@ -328,52 +328,55 @@ func (c *Call) matches(args []interface{}) error {
...
-			if !m.Matches(args[i]) {
-				return fmt.Errorf("Expected call at %s doesn't match the argument at index %s.\nGot: %v\nWant: %v",
-					c.origin, strconv.Itoa(i), args[i], m)
+			// The number of actual args don't match the number of matchers,
+			// or the last matcher is a slice and the last arg is not.
+			// If this function still matches it is because the last matcher
+			// matches all the remaining arguments or the lack of any.
+			// Convert the remaining arguments, if any, into a slice of the
+			// expected type.
+			vargsType := c.methodType.In(c.methodType.NumIn() - 1)
```
balshetzer: I think we need a check before here that the function is variadic. Otherwise, we could get a panic when trying to call MakeSlice on a non-slice type. Also, we probably shouldn't try to match the wrong number of args in a non-variadic function.

gomock/controller_test.go
aaronbee: To reproduce the original panic I needed a func that took in an argument of a different type than the vararg argument.

gomock/call.go
balshetzer: Right, of course.

PR #132, opened 2017-11-21
add AssignableTo(..) to matchers

josh-tepper:
A fairly generic matcher that I've found useful -- might be worth committing to the library

### DIFF
--- a/gomock/matchers.go
+++ b/gomock/matchers.go
@@ -103,6 +103,7 @@ func (m assignableToMatcher) String() string {
 func Any() Matcher             { return anyMatcher{} }
 func Eq(x interface{}) Matcher { return eqMatcher{x} }
 func Nil() Matcher             { return nilMatcher{} }
+
 // May pass either a value or another Matcher
 func Not(x interface{}) Matcher {
 	if m, ok := x.(Matcher); ok {


gomock/matchers.go
```diff
@@ -87,13 +87,30 @@ func (n notMatcher) String() string {
...
 func Nil() Matcher             { return nilMatcher{} }
+
+// May pass either a value or another Matcher
 func Not(x interface{}) Matcher {
 	if m, ok := x.(Matcher); ok {
 		return notMatcher{m}
 	}
 	return notMatcher{Eq(x)}
 }
+func AssignableTo(t interface{}) Matcher {
```
balshetzer: How is this called? Maybe an example in the comment would help. Also, would you mind adding a unit test?

pasztorpisti:
To me it feels a bit narrow purpose - most tests don't need something like this.

msabramo:
I find this very useful when mocking functions that take `interface{}` parameters.

For example, let's say I need to have a mock for this interface:

```go
type Inserter interface {
    Insert(list ...interface{}) error
}
```

Using this matcher lets me take:

```go
	dbMock.EXPECT().
		Insert(gomock.Any()).
		Return(errors.New("DB error"))
```

and change it go:

```go
	dbMock.EXPECT().
		Insert(gomock.AssignableToTypeOf(&EmployeeRecord{})).
		Return(errors.New("DB error"))
```

which makes the test much more readable because I can see at a glance what is being inserted, which could be important if the code under test does multiple inserts to different tables. Also if I am expecting the mock to be called with a certain type of parameter and it gets called with a different type, the test failure will let me know right away that my assumption was wrong.

In fact I ended up implementing this independently, but then went back and borrowed some things from this PR to make a new PR with unit tests and documentation examples: 

https://github.com/golang/mock/pull/140

pasztorpisti:
@msabramo In such situations I always expected an exact match. I'm wondering in which scenario it is sensible to check only the type without actual value. To me it feels like making the test less strict/robust.

msabramo:
@pasztorpisti: Yeah that is probably true in some situations, but in my case one of the structs in the system I'm working on has 41 fields, so I think having tests that assert the value of those 41 fields will be overly verbose and also overly brittle. Also some of those fields are timestamp fields, so I don't even think it's possible for me to fully specify the struct and have it match.

I think your exact match idea would work in a lot of cases where the structs are simple, but when it comes to structs that are large and/or have dynamic fields like timestamps, random UUIDs, etc., I don't think exactness is practical and `AssignableToTypeOf` allows me to be much stricter than `gomock.Any()` but not quite as strict as `gomock.Eq`.

PR #133, opened 2017-11-28
Fatal on duplicate Controller.Finish call

pasztorpisti:
Failing with `t.Fatal` in case of multiple `Finish` calls on the same controller instance.

This would prevent sharing the same controller between subtests.

### DIFF
--- a/gomock/controller.go
+++ b/gomock/controller.go
@@ -77,6 +77,7 @@ type Controller struct {
 	mu            sync.Mutex
 	t             TestReporter
 	expectedCalls *callSet
+	finished      bool
 }
 
 func NewController(t TestReporter) *Controller {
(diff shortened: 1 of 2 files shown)


gomock/controller.go
```diff
@@ -182,6 +183,11 @@ func (ctrl *Controller) Finish() {
 	ctrl.mu.Lock()
 	defer ctrl.mu.Unlock()
 
+	if ctrl.finished {
+		ctrl.t.Fatalf("Controller.Finish has to be called exactly once")
```
balshetzer: I'd like the error message to tell the user what went wrong in addition to explaining why it was wrong. For example:
"Controller.Finish was called more than once. It has to be called exactly once."

pasztorpisti: Done.

Issue #85, opened 2017-06-27
Source mode fails with embedded interfaces in a specific case


pasztorpisti:
Here are my test files (2 packages):

*github.com/anyuser/pkg1/pkg1.go:*
```go
package pkg1

type I interface {
	E
}

type E interface {
	M()
}
```

*github.com/anyuser/pkg2/pkg2.go:*
```go
//go:generate mockgen -source pkg2.go -destination pkg2_mock.go -package pkg2 -aux_files pkg1=../pkg1/pkg1.go
package pkg2

import "github.com/pasztorpisti/pkg1"

type I interface {
	pkg1.I
}
```

If you run `go generate` on `pkg2` then you get the following error message:
```
2017/06/27 18:03:01 Loading input failed: ../pkg1/pkg1.go:4:2: unknown embedded interface E
pkg2.go:1: running "mockgen": exit status 1
```

Note that mock generation succeeds if you go to `pkg1` and embed the `E` interface into `I` by prefixing it with the package name `pkg1.E` but that turns `pkg1.go` into invalid go code that doesn't compile.

JeremyLoy:
This also works in the following scenario:

_github.com/anyuser/pkg1/file1.go_
`package pkg1

type A interface {
	Method1()
}
`

_github.com/anyuser/pkg1/file2.go_
`package pkg1

type B interface {
	A
}
`

Just to make explicit the difference here: Two separate files, but same package produces the same issue.

JeremyLoy:
Looking into it more, I found this TODO in the source code.

https://github.com/golang/mock/blob/747e9e6390e12b0d987c261bdc89a51e76b9f5fc/mockgen/mockgen.go#L18

balshetzer:
Both given examples above work now. It's time to remove that TODO.

amatoenot:
@balshetzer Just checked. Not working. The only way to solve this case is to use reflect mode.

PR #134, opened 2017-11-28
Fix resolution of embedded interfaces in aux files

notbdu:
Embedded interfaces in `aux_files` generate `unknown embedded interface XXX` errors.
See below for example of the problem:
```
// source
import (
    alias "some.org/package/imported"
)

type Source interface {
    alias.Foreign
}
```

```
// some.org/package/imported
type Foreign interface {
    Embedded
}

type Embedded interface {}
```

Attempting to generate a mock will result in an `unknown embedded interface Embedded`.
The issue is that the `fileParser` stores `auxInterfaces` underneath the package name
explicitly specified in the `aux_files` flag.

In the `parseInterface` method, there is an incorrect assumption about an embedded interface
always being in the source file.
```
case *ast.Ident:
        // Embedded interface in this package.
        ei := p.auxInterfaces[""][v.String()]
        if ei == nil {
                return nil, p.errorf(v.Pos(), "unknown embedded interface %s", v.String())
        }
```

Test case verifies that the patch fixes the problem described in the bug report.


pasztorpisti:
seems to be related to https://github.com/golang/mock/issues/85

notbdu:
@pasztorpisti thanks for referencing those issues

wsong:
This PR breaks this repo on Windows; `aux` is not a valid directory name in Windows (https://social.technet.microsoft.com/Forums/Azure/en-US/d1cf7394-a39b-46bf-823d-46f33b2f78d0/unable-to-create-folder-named-aux-anywhere-in-the-disk?forum=w7itprogeneral), so this PR prevents you from checking out this repo on a Windows machine.

notbdu:
@wsong thanks for bringing it up. Opened a second PR w/ `aux` renamed to `faux`.

PR #135, opened 2017-12-06
Rename aux -> faux

notbdu:
- `aux` is not a valid directory name on windows
- Fixes https://github.com/golang/mock/pull/134#issuecomment-349758272


notbdu:
@balshetzer sorry to bother, but someone just pointed out that `aux` is not a valid directory name on Windows. Just fixed the issue here by renaming `aux` to `faux`.

marten-seemann:
This should be merged ASAP.

PR #140, opened 2018-01-03
New matcher: gomock.AssignableToTypeOf

msabramo:
This is the exact same idea as https://github.com/golang/mock/pull/132 from @josh-tepper, because I forgot to check for an existing PR for this feature before writing the code (!). But then I took some of the ideas from that PR and merged them with my code to create this PR, which has unit tests and documentation. So this should be a considered a team effort from @josh-tepper and me.

This is very useful when mocking functions that take `interface{}` parameters.

For example, let's say I need to have a mock for this interface:

```go
type Inserter interface {
    Insert(list ...interface{}) error
}
```

Using this matcher lets me take:

```go
	dbMock.EXPECT().
		Insert(gomock.Any()).
		Return(errors.New("DB error"))
```

and change it go:

```go
	dbMock.EXPECT().
		Insert(gomock.AssignableToTypeOf(&EmployeeRecord{})).
		Return(errors.New("DB error"))
```

which makes the test much more readable because I can see at a glance what is being inserted, which could be important if the code under test does multiple inserts to different tables. Also if I am expecting the mock to be called with a certain type of parameter and it gets called with a different type, the test failure will let me know right away that my assumption was wrong.

Cc: @balshetzer, @seanisom, @iderdik, @shylu, @abarla2010

### DIFF
--- a/gomock/matchers.go
+++ b/gomock/matchers.go
@@ -87,6 +87,18 @@ func (n notMatcher) String() string {
 	return "not(" + n.m.String() + ")"
 }
 
+type assignableToTypeOfMatcher struct {
+	targetType reflect.Type
+}
+
+func (m assignableToTypeOfMatcher) Matches(x interface{}) bool {
+	return reflect.TypeOf(x).AssignableTo(m.targetType)
+}
+
+func (m assignableToTypeOfMatcher) String() string {
+	return "is assignable to " + m.targetType.Name()
+}
+
 // Constructors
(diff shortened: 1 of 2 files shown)


msabramo:
gentle ping

msabramo:
ping

msabramo:
Great! Thanks, @balshetzer!

letientai299:
Hi, this get merged for a long time already, and is a very useful feature, so why don't we bump release version to include this feature?

PR #148, opened 2018-02-07
Fix bug with comparing variadic arguments

ppg:
Introduced in either 7dc66f5f789eca1f66e82966331cde181184e62e or f159d62078cd1ee294e58036af3e8809ec88de8b, I can't tell which one.

The current code with variadic arguments on a mock will compare the first of those arguments, but then it breaks and doesn't compare the rest. This leads to both false equivalency, and the side effect of returning the wrong values from a partially matched mock.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,5 +13,6 @@
 # The mockgen binary.
 mockgen/mockgen
 
-# A binary produced by gotest.
+# Binaries produced by gotest.
 #gomock/[568]\.out
+*.test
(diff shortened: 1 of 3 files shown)


johandroz:
Hello, 

I am interested in this PR, I experienced the same issue with variadic arguments.
Any update on the merge?

Thanks in advance,
Johan

balshetzer:
The author needs to sign the CLA before the PR can be committed.

balshetzer:
I addressed this issue with #165

PR #149, opened 2018-02-08
support argument capture matcher

trung:
Allows capturing arguments that are passed to the mocked function call

```
argumentCaptor := gomock.CaptureArgument().(*gomock.ArgumentCaptor)
someMock.EXPECT().SomeMethod(argumentCaptor).Return(abc).Time(2)

foo.Bar() // internally call someMock.SomeMethod(...)

capturedArguments := argumentCaptor.GetAll() // slice contains 2 as .Time(2) above
// or
singleArgument := argumentCaptor.GetSingle() // first value
```

### DIFF
--- a/gomock/matchers.go
+++ b/gomock/matchers.go
@@ -87,6 +87,29 @@ func (n notMatcher) String() string {
 	return "not(" + n.m.String() + ")"
 }
 
+type ArgumentCaptor struct {
+	values []interface{}
+}
+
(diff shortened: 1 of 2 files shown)


poy:
@trung Thanks for the PR. Could you not achieve the same thing with `Do(...)`:

var capturedArgs []int

```go
someMock.
  EXPECT().
  SomeMethod(gomock.Any()).
  Do(func(arg int){
    capturedArgs = append(capturedArgs, arg)
  })
```

PR #151, opened 2018-02-08
Extract parser and reflect to packages

dnovikoff:
The ParseFile and Relfect methods with the model package provides some useful and generic functionality, that could be reused by other parsers/generators.

Looks like that would be cool to extract them to separate packages


dnovikoff:
I signed it!

poy:
@dnovikoff Thanks for the PR. 

I don't think we want to support the parsing/reflect as an exported functionality.

Issue #152, opened 2018-02-15
Mockgen: generate interface mock in another package when method depends from another types of the same package


nawa:
sample project structure is

```
gomocktst
├── mock_data
│   └── interface_mock.go
├── interface.go
├── interface_test.go
```

`interface.go` contains code:

```go
package gomocktst

type ReturnType string
type ParamType string

type Interface interface {
	Method(param ParamType) ReturnType
}
```

I want to generate mock of `Interface` in the `mock_data` package with next command

```
mockgen -source interface.go -package mock_data -destination mock_data/interface_mock.go
```

As a result I get inconsistent `MockInterface` where `ParamType` and `ReturnType` aren't imported. More detailed example here https://github.com/nawa/gomocktst/blob/master/mock_data/interface_mock.go#L36

Is it a bug, or I do something wrong?

vail130:
I'm having this same issue. Would love some guidance on a workaround or status of a fix.

nawa:
@vail130 The workaround is to fix it manually

vail130:
I opened a PR with a fix, and then found another PR already opened with a fix. And then I switched to pegomock, which works :)

balshetzer:
This has been fixed by #164.

PR #154, opened 2018-02-22
Adds support for exported types in interface signatures

vail130:
Fixes https://github.com/golang/mock/issues/152


vail130:
I signed it!

mockgen/mockgen.go
```diff
@@ -42,6 +42,7 @@ const (
 
 var (
 	source          = flag.String("source", "", "(source mode) Input Go source file; enables source mode.")
+	sourcePackage   = flag.String("source_package", "", "Full package name for interface to be mocked. Required to support references to structs from same package.")
```
balshetzer: Can this be figured out automatically from the source rather than needing to be specified on the command line?

mockgen/tests/no_import_for_structs_in_interface/README.md
```diff
@@ -0,0 +1,35 @@
+Types that appear in the signature of a mocked interface are not imported
```
balshetzer: This test case is phrased as a bug report. It will start to make less sense when someone reads it in the future. Rather than word it as "x is not working" you could state it as "verify that x works correctly".

balshetzer:
The issue being addressed here was fixed by #164.

Issue #31, opened 2016-04-12
reflect mode failed while model.InterfaceFromInterfaceType print to os.stdout


rantibi:
I run the following command :`mockgen matching  BannedDA` 
since it used reflect mode, it generate reflectProgram witch write the pkg to os.stdout, and then the mockgen try to read the resuld and decode it.

since I have init() method that print to std (using logrus logger) in part of my files, the stdout is not clean, and therefor the pkg can't be decoded in the following [line](https://github.com/golang/mock/blob/master/mockgen/reflect.go#L96), and the mockgen failed

I try to debug, and found that is [`{"BannedDA", reflect.TypeOf((*pkg_.BannedDA)(nil)).Elem()}`](https://github.com/golang/mock/blob/master/mockgen/reflect.go#L131) cause the init() call.

The point is the we can't trust stdout, and assume that it contains only the encode pkg,

dsymonds:
Random packages should not be printing to stdout at init time. Log messages should be going to stderr anyway, but it's still obnoxious for packages to spam a program's output streams like that.

Fix your code and it should work.

rantibi:
Well thats mean that if I use some third party (which I can't change) the write to stdout I can't use mockgen?

dsymonds:
I guess so, yes. Such code would not be suitable in lots of circumstances.

rantibi:
why not add some support for that?  (I can do it and open PR for that)
Thats mean I will print some result mark line before this [line](https://github.com/golang/mock/blob/master/mockgen/reflect.go#L150)
something like this `fmt.Println('\nPKG_RESULT')`
then [here](https://github.com/golang/mock/blob/master/mockgen/reflect.go#L93) I will read from then stdout until I find this mark line, and only then try to do the decoding.

what do you say?

(If not... I will have to fork it, which is something that really don't want to do)

dsymonds:
Sorry, it's something that just isn't worth trying to work around. Libraries shouldn't be writing to stdout. File a bug against the library that's doing it.

PR #175, opened 2018-04-15
Ignore stdout does not matter when decoding pkg content

yutianwu:
Hi, I have encountered the same problem described in https://github.com/golang/mock/issues/31. 

When I use gomock to generate code and if there is anything else printed to stdout, the program will complain "Loading input failed: extra data in buffer". 

IMO, Gomock is the best mock tool so far, and I do need it urgently. But I cannot use it when some other pkgs print stuff to stdout. I have done a little modification to ignore stdout that I don't need. 

I can use it now but I will appreciate it if you can merge my pr then I don't need to maintain my fork any longer. 

thx for you time~


yutianwu:
I signed

yutianwu:
I updated my profile

yutianwu:
Done

balshetzer:
@yutianwu 

I made some changes and posted it at #176. Could you please post on #176 that you consent to submitting this code? 

Thanks.

balshetzer:
The purpose of this change is to handle the case where something is written to stdout on init(). Nothing else in the program is run besides the init.

If the package starts goroutines on init() which randomly write to stdout then we are in trouble. Writing to stdout is not synchronized so trying to write an entire buffer at once or putting sentinels before and after the content will not protect us from pollution.

yutianwu:
Agreed, but I did encountered such a problem that goroutines of other packages written something to stdout after we print the indicator because we need a little time to generate pkg content and it happens all the time.
We cannot solve all problems with two indicators as you said but it helps sometimes.

balshetzer:
If that's the case then I'd rather have the program write the output to a file in the temporary directory and read the output from there. This should always work.

yutianwu:
Yeah, I thought about that. If you agree to do that, I'd be glad to do it and then there will be no problems.

yutianwu:
@balshetzer 
Hi, I'm glad that you agree to write pkg content to temporary file, I have done some changes and could you please review it. Thank you very much

balshetzer:
I addressed this issue with #184

PR #176, opened 2018-04-16
Handle stdout pollution

balshetzer:
Allow gomock reflect to work even when there is an init() that writes to stdout.


yutianwu:
Hi, the case I ran into is one goroutine print some log when init. and the result is like
```
GOMOCK_PKG_CONTENT_START_INDICATOR
INFO ...
package ..
```
You can test it if you start a goroutine in init() and sleep 1s then print some stuff and after print indicator then sleep 2s. you will get the result like I got.

yutianwu:
hey buddy, I have committed my change in my pr. 

could you pls review it, thx a lot. I really need it.

balshetzer:
I addressed this issue with #184

PR #177, opened 2018-04-18
Add Yield and YieldValues methods

msabramo:


poy:
@msabramo Thanks for the PR and sorry for the slow response.

This feels like similar behavior to what you can do with [Call.Do](https://godoc.org/github.com/golang/mock/gomock#Call.Do) but maybe a bit more complicated due to the go-routines and channels.

msabramo:
OK, I'll close this then.

PR #179, opened 2018-04-19
Add IsOfType matcher

msabramo:


poy:
This seems very similar to the `AssignableToTypeOf` matcher. Do we need both?

msabramo:
Yeah they have similar purposes. I would point out that `IsOfType` has a more intuitive name and semantics, because you don't have to create an instance of a type just to test that something is of the same type.

Compare:

```go
walkerMock.EXPECT().Walk(gomock.AssignableToTypeOf(&Dog{}))
```

vs.

```go
walkerMock.EXPECT().Walk(gomock.IsOfType("Dog"))
```

poy:
I'm not so sure about this one. I think its a neat concept but I also think it introduces quite a few edge cases. To name a few:

- This doesn't support types from other packages:
```
r := &http.Request{}
fmt.Println(reflect.TypeOf(r).Name()) // This doesn't output anything
```
- I could be silly and make a `nil` type and trick this:
```
type nil struct{}
n := nil{}
fmt.Println(reflect.TypeOf(n).Name())

// Output: nil
```

I'm sure these could be worked around and sorted out, however I'm nervous about introducing the complexity. 

Perhaps this would be better suited in its own library somehow. If we had a matcher that took a `reflect.Type`, perhaps there could be a library that had a function that went from `string` to `reflect.Type` that could work with the aforementioned matcher well.

I'm going to mark this as wontfix for now.

PR #180, opened 2018-04-19
Add ImplementsInterface matcher

msabramo:


msabramo:
.

poy:
@msabramo This seems very related to #179 Is it necessary to have such similar matchers?

scevallos:
I agree that there's no need for an "IsOfType" matcher (from #179) since it is very similar to the existing AssignableToTypeOf. However, adding an ImplementsInterface Matcher could be useful and not something currently encapsulated by the existing AssignableToTypeOf.

msabramo:
Yeah, `IsOfType` is very similar to `AssignableToTypeOf`. The difference is syntactical but the capabilities are the same. But with `ImplementsInterface`, that's different. It's much looser, since a lot of different types can satisfy an interface.

poy:
I've often seen the convention:
```go

type SomeInterface interface {
  Foo()
}

type SomeType struct{}

var _ SomeInterface = SomeType{} // <- Asserts that type implements interface
```

Is there a reason adding a matcher would work better than just letting the compiler do its thing?

codyoss:
I am closing this for now due to lack of response. Please reopen if you are interested in making the updates. I also think the language provides good ways of checking this without using a library as stated in the comment above.

PR #189, opened 2018-05-23
Implement Matchers for "loose" variadic argument matching: All, Len, Contains

jinnovation:
This PR provides a middle-ground approach to recording variadic argument conditions, between simply using `gomock.Any()` (accepting any list of argument values) or matching against exact argument lists. To that effect, we provide two Matchers to allow for "looser" matching of variadic parameters, including:
* `Contains`, which allows you to associate expected mock behavior with variadic argument lists simply containing one or more values; and
* `Len`, which likewise allows you to associate with lists of certain length.

This PR also provides an `All` compound Matcher, which, for example, can be used—in conjunction with both `Contains` and `Len`—to match against unordered values of a variadic argument slice.

### DIFF
--- a/gomock/matchers.go
+++ b/gomock/matchers.go
@@ -19,6 +19,7 @@ package gomock
 import (
 	"fmt"
 	"reflect"
+	"strings"
 )
 
 // A Matcher is a representation of a class of values.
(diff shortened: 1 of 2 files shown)


gomock/matchers.go
```diff
@@ -122,3 +123,86 @@ func Not(x interface{}) Matcher {
...
+	return reflect.TypeOf(x).Kind() == reflect.Slice &&
+		m.n == reflect.ValueOf(x).Cap()
+}
+
+func (m lenMatcher) String() string {
+	return fmt.Sprintf("is of length %v", m.n)
+}
+
+// Len returns a Matcher that matches on length. The returned Matcher returns
+// false if its input is not a slice.
```
poy: Could we also add support for chan and map?

gomock/matchers.go
```diff
@@ -122,3 +123,86 @@ func Not(x interface{}) Matcher {
...
+	return compositeMatcher{ms}
+}
+
+type lenMatcher struct {
+	n int
+}
+
+func (m lenMatcher) Matches(x interface{}) bool {
+	return reflect.TypeOf(x).Kind() == reflect.Slice &&
+		m.n == reflect.ValueOf(x).Cap()
```
poy: Should we be using `.Len()` instead of `.Cap()`?

gomock/matchers.go
```diff
@@ -122,3 +123,86 @@ func Not(x interface{}) Matcher {
 func AssignableToTypeOf(x interface{}) Matcher {
 	return assignableToTypeOfMatcher{reflect.TypeOf(x)}
 }
+
+type compositeMatcher struct {
```
poy: nit: Can we rename this to `allMatcher` to better match the existing convention?

gomock/matchers.go
```diff
@@ -122,3 +123,86 @@ func Not(x interface{}) Matcher {
...
+
+// Contains returns a Matcher that asserts that its input contains the requisite
+// elements, in any order. The returned Matcher returns false if its input is
+// not a slice. It is useful for variadic parameters where you only care that
+// specific values are represented, not their specific ordering.
+func Contains(vs ...interface{}) Matcher {
+	return containsElementsMatcher{vs}
+}
+
+type containsElementsMatcher struct {
```
poy: nit: Can we rename this to `containsMatcher` to match existing convention?

gomock/matchers.go
```diff
@@ -122,3 +123,86 @@ func Not(x interface{}) Matcher {
...
+		return false
+	}
+
+	xv := reflect.ValueOf(x)
+
+	eq := true
+Elements:
+	for i := 0; i < len(m.Elements) && eq; i++ {
+		e := m.Elements[i]
+		for j := 0; j < xv.Cap(); j++ {
```
poy: Should this be using `Len()` instead of `Cap()`?

gomock/matchers.go
```diff
@@ -122,3 +123,86 @@ func Not(x interface{}) Matcher {
...
+
+func (m containsElementsMatcher) Matches(x interface{}) bool {
+	if reflect.TypeOf(x).Kind() != reflect.Slice {
+		return false
+	}
+
+	xv := reflect.ValueOf(x)
+
+	eq := true
+Elements:
```
poy: nit: for readability, WDYTA putting the inner loop in a helper function and removing the label?

gomock/matchers.go
```diff
@@ -122,3 +123,86 @@ func Not(x interface{}) Matcher {
...
+	for i := 0; i < len(m.Elements) && eq; i++ {
+		e := m.Elements[i]
+		for j := 0; j < xv.Cap(); j++ {
+			if reflect.DeepEqual(xv.Index(j).Interface(), e) {
+				continue Elements
+			}
+		}
+		eq = false
```
poy: nit: for readability, `return false` here and remove the condition above?

codyoss:
@jinnovation are you still interested in this PR?

jinnovation:
@codyoss: Yes, I am. Apologies about the delay; I should be able to address @poy's feedback in the near future.

codyoss:
Awesome, thanks for your PR!

codyoss:
@jinnovation Still interested?

jinnovation:
@codyoss: Apologies for leading your team on. Despite expectations, I unfortunately haven't had a chance to follow up on this PR. Feel free to close this out; I will re-open if I come around to it.

codyoss:
I might pick this up in that case as I think these features would be nice to have.

codyoss:
I am going to close this for now. But I will probably open up individual PRs for these style of matchers in the future. Thank you for your work on this though, we will try to get a similar feature set it!

PR #195, opened 2018-07-06
Create the path at which mock will be generated

theherk:
When the path does not exist an unnecessary failure is generated. Creating the
directories required is harmless and simplifies generation.


theherk:
I signed it!

PR #196, opened 2018-07-19
When searches for a matching call, return most recently expected call.

tocrafty:
Most recently added action will run after function is called. When this action is exhausted, old action will be triggered

### DIFF
--- a/gomock/callset.go
+++ b/gomock/callset.go
@@ -69,12 +69,12 @@ func (cs callSet) FindMatch(receiver interface{}, method string, args []interfac
 	// Search through the expected calls.
 	expected := cs.expected[key]
 	var callsErrors bytes.Buffer
-	for _, call := range expected {
-		err := call.matches(args)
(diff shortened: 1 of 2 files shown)


tocrafty:
I signed it!

balshetzer:
Could you please explain why it's better to use the latest action rather than the order in which they were added?

tocrafty:
Sometimes, we need to set up default mock behaviors before test suite. In a test case, we can set up new actions for interface if we want to simulate some erros. After this actions is exhausted, mock behavior will fall back to default setup. With old strategy, all actions must be add as planed and carefully, And they can not be changed after setup. This would be very inflexible.

poy:
If I'm understanding correctly, this is an attempt to add support for loose mocks in a way. Is that correct?

If so, I've been pointing people towards #246 for a discussion there.

tocrafty:
Yeah, you are right. I have read read #238 and #246 , and this PR should be redirected to #238 as an mock options for backward compatibility. I will close this one.

PR #208, opened 2018-08-30
mockgen package name replaced by main to mockgen

otokaze:


otokaze:
retry

otokaze:
retry

otokaze:
retry

balshetzer:
Why are you proposing this change?

otokaze:
Thank you for your reply.
Because your mockgen tool only supports single file mock code generation in source mode.
This design is not very friendly to my project and can't be well integrated with it, so we decided to fork a copy of your code and plan to change it. mockgen package name replaced by main to mockgen，mainly supports the following use.
Such as

func genMock(files ...string) (err error) {
	for _, file := range files {
		var pkg *model.Package
		if pkg, err = mockgen.ParseFile(file); err != nil {
			return
		}
		if len(pkg.Interfaces) == 0 {
			continue
		}
		var filename = pkg.SrcDir + "/" + pkg.Name + "_mock.go"
		if _, err := os.Stat(filename); !os.IsNotExist(err) {
			continue
		}
		var g = &mockgen.Generator{Filename: file}
		if err = g.Generate(pkg, pkg.Name, filename); err != nil {
			return
		}
		if err = ioutil.WriteFile(filename, g.Output(), 0644); err != nil {
			return
		}
	}
	return
}

PR #210, opened 2018-09-11
Store the source hash to avoid generating content for unmodified files

rafaeljusto:
Keeping the hash of the source in the generated file allows to bypass the generation when the source wasn't modified.


romanoaugusto88:
:tada: Suggestion: add a `-cache` flag to be optional this behaviour.

balshetzer:
Can you explain the benefits of this change?

rafaeljusto:
@balshetzer Sure! I will explain the scenario that this change would be helpful.

We currently use `mockgen` to generate all interfaces mocks of our project, executing manually the commands to update/create a mock every time an interface is modified/created. All mock files are added to the repository.

Our plans now are to generate automatically the mocks adding the `mockgen` command to a `go:generate` entry where the interface is located. The mocks are still being added to the repository, but now we don't need to manually run the commands anymore.

### The problem

When running `mockgen` in `go generate ./...` it will process all files again, even those interfaces that weren't modified. This slows down a lot the build as the file parsing of `mockgen` takes some time. Checkout the difference:

```
Without mockgen:
go generate ./...  0.68s user 0.92s system 96% cpu 1.667 total

With mockgen:
go generate ./...  36.66s user 11.90s system 131% cpu 37.012 total
```

### Solution

My solution was to store the hash of the source file in the generated file (as a comment), so we can compare the hashes before doing any parsing.

rafaeljusto:
I just realised that the commands were slow because of the reflect mode. When changing to source mode this becomes really fast, removing the necessity of a cache solution.

PR #220, opened 2018-10-19
fixed #156 context import collision

### LINKED ISSUE #156: handle golang.org/x/net/context & context import collisions with source mode
So I've just been testing mock with protocol buffer generated code on a simple .proto file

```
syntax = "proto3";

package pb;

// A generic empty message
message Empty {

}

// The Add service definition.
service Random {
    // Gets a random number.
    rpc Get (Empty) returns (stream GetReply) {}
}

// The get response contains the random number.
message GetReply {
    int64 v = 1;
    string err = 2;
}

```

To generate the above you can do 

Install proto3
 https://github.com/google/protobuf/releases
 mv protoc/bin/protoc /usr/local/bin/protoc

Update protoc Go bindings via
 go get -u github.com/golang/protobuf/{proto,protoc-gen-go}

`
protoc random.pb --go_out=plugins=grpc:.;
`

The important part to notice of this definition is the usage of streams which causes an interface of interfaces to get generated by protoc

`type Random_GetClient interface {
	Recv() (*GetReply, error)
	grpc.ClientStream
}
`

When using mockgen on this file I get "imported package collision: context imported twice".
Now I understand the need for an error here because its possible to have a naming collision. In this case however the problem is because golang.org/x/net/context & context paths are actually the same package. 

I'm not sure of a way to test two imports are the same but I made a less generic fix here: https://github.com/Rajczyk/mock

linuxuser586:
This addresses the issue where a package imports another package with context
and golang.org/x/net/context in the same included package. This is leagal since
golang.org/x/net/context is an alias for context. We still don't want other
aliases to context to pass. There are tests that varify that only context and
golang.org/x/net/context are compatable for this special case.

### DIFF
--- a/mockgen/parse.go
+++ b/mockgen/parse.go
@@ -35,6 +35,8 @@ import (
 var (
 	imports  = flag.String("imports", "", "(source mode) Comma-separated name=path pairs of explicit imports to use.")
 	auxFiles = flag.String("aux_files", "", "(source mode) Comma-separated pkg=path pairs of auxiliary Go source files.")
+	// for easier testing
+	fatalf = log.Fatalf
 )
 
 // TODO: simplify error reporting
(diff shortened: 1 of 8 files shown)


poy:
I would prefer we not add these special cases. I would also hope that the context stuff is resolved at this point.

I'm going to close this for now, feel free to reopen it.

PR #227, opened 2018-11-21
Fatal can only be called once

poy:
It is invalid to call Fatal more than once. Finish currently will call Fatal again in some cases.


codyoss:
Closing this PR as I believe the changes or no longer relevant with some of the more recent updates around supporting Go 1.14 cleanup method.

PR #228, opened 2018-11-21
remove full path from callerInfo

poy:
The test package provides enough context on where the file is located.


poy:
@balshetzer This one probably needs some discussion. I personally think having the full path to the file is clutter and obfuscates the results. Thoughts?

codyoss:
I am going to close this PR for now. If you are still interested in the changes lets open up an issue and talk about it first. Thanks.

PR #229, opened 2018-11-22
Resolves #147 - Including binaries with tagged releases via GoReleaser

### LINKED ISSUE #147: Include mockgen binaries with tagged releases
It would be really helpful to include `mockgen` binaries with tagged releases similar to [glide](https://github.com/Masterminds/glide/releases). This makes it easier for using the correct version of `mockgen` to generate mocks for a specific version of `gomock`.

scevallos:
[GoReleaser](https://goreleaser.com) provides an easy way to automate releases via a single config YAML. You can see what the result would look like on my [fork's releases](https://github.com/scevallos/mock/releases). The maintainers of this repo would have to generate a valid GitHub Token and set it as an envvar as specified [here](https://goreleaser.com/environment/). This step can also be [integrated with Travis](https://goreleaser.com/ci/), if the maintainers want to do that.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,39 @@
+# Release automation via GoReleaser (goreleaser.com)
+# Requires a valid GITHUB_TOKEN envar prior to running `goreleaser`
+# See https://goreleaser.com/environment/ for more info
+---
+release:
+  github:
+    owner: golang
+    name: mock
+
+builds:
+  - binary: mockgen
+    goos:
+      - darwin
+      - windows
+      - linux
+    goarch:
+      - amd64
+      - 386
+    env:
+      - CGO_ENABLED=0
+    main: ./mockgen/
+archive:
+  format: tar.gz
+  wrap_in_directory: true
+  name_template: '{{ .Binary }}-{{ .Version }}-{{ .Os }}-{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}'
+  files:
+    - LICENSE
+    - README.md
+checksum:
+  name_template: '{{ .Binary }}-{{ .Version }}-checksums.txt'
+snapshot:
+  name_template: "snap-{{ .Commit }}"
+changelog:
+  sort: asc
+  filters:
+    exclude:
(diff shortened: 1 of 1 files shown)


scevallos:
I signed it!

.goreleaser.yml
```diff
@@ -0,0 +1,39 @@
...
+    goarch:
+      - amd64
+      - 386
+    env:
+      - CGO_ENABLED=0
+    main: ./mockgen/
+archive:
+  format: tar.gz
+  wrap_in_directory: true
+  name_template: '{{ .Binary }}-{{ .Version }}-{{ .Os }}-{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}'
```
poy: What does `.Arm` correspond to and why do we want the `v` suffix?

scevallos: `.Arm` corresponds to the value of GOARM as specified on [this page](https://github.com/golang/go/wiki/GoArm) and is used to distinguish between the different versions. If you don't wish to support ARM, this can be removed.

poy: Ah, perfect! I was missing those docs. Thanks for the clarification and I see no reason to exclude ARM.

kishaningithub:
@poy @scevallos I am also thinking if we could also use [godownloader](https://github.com/goreleaser/godownloader) installation would become totally cool. This tool generates an installation script that you can check in into this repo. This installation script takes care of detecting the host platform and installing the appropriate binary. An example -  [gosec](https://github.com/securego/gosec#ci-installation)

codyoss:
Fixes: #147

.goreleaser.yml
```diff
@@ -0,0 +1,39 @@
...
+builds:
+  - binary: mockgen
+    goos:
+      - darwin
+      - windows
+      - linux
+    goarch:
+      - amd64
+      - 386
+    env:
```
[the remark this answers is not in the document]
scevallos: Done!

codyoss:
LGTM, Thanks for your contribution @scevallos 🎆

PR #235, opened 2018-11-29
Add support for generating mocks for simple callback types.

EdSchouten:
Right now mockgen only emits mocks for interface types it encounters.
There are various cases, even in the standard library, where function
types are declared for simple callback purposes. These are currently
only mockable by adding a custom interface.

This change extends both the parser and the reflection code to
automatically convert function types to interfaces, having a single
"Call" method.


balshetzer:
I don't understand the problem that this fixes. Please show a simple but complete example of a system that you might want to test that has this problem and how you want gomock to be used to test.

Also, instead of modifying an existing test to include this use case please make a new test for it. Ideally, the test would be the illustrative example above.

EdSchouten:
Anything else I could do from my side to get this integrated?

EdSchouten:
Ping?

EdSchouten:
Just brought this fix in sync with latest master again.

EdSchouten:
Friendly ping!

mockgen/tests/plain_func/input.go
```diff
@@ -0,0 +1,5 @@
+//go:generate mockgen -destination reflect_output/mock.go github.com/golang/mock/mockgen/tests/plain_func Func
+//go:generate mockgen -source input.go -package plain_func -destination source_output/mock.go
+package plain_func
+
+type Func func(int) int
```
poy: Can we add some code that uses the generated output? This should help to ensure we don't break this down the road?

mockgen/parse.go
```diff
@@ -512,6 +526,38 @@ func iterInterfaces(file *ast.File) <-chan namedInterface {
...
 
+type namedFunc struct {
+	name *ast.Ident
+	ft   *ast.FuncType
+}
+
+// Create an iterator over all function types in file.
+func iterFuncs(file *ast.File) <-chan namedFunc {
+	ch := make(chan namedFunc)
+	go func() {
```
poy: Can we return a slice instead of a channel? I don't think we need the extra complexity of a go routine and channels here.

EdSchouten: Though I agree returning a slice would also be acceptable here, this was done to be consistent with `iterInterfaces()` (placed right above `iterFuncs()`).

Would you want me to convert both of these to using a slice? I guess we should also rename these functions then, as they then don't iterate. `getInterfacesInFile()` and `getFuncsInFile()`?

mockgen/tests/plain_func/input.go
EdSchouten: I just pushed a second commit to add such a test. While there, I also moved the test to the right location. It looks like the tests were moved around at some point.

mockgen/parse.go
poy: I think I would prefer we improve the codebase instead of continuing questionable patterns. WDYTA a refactor commit?

EdSchouten: Done!

codyoss:
@EdSchouten could you rebase with master, I am going to start reviewing a bunch of PRs this weekend

EdSchouten:
> @EdSchouten could you rebase with master, I am going to start reviewing a bunch of PRs this weekend

Done!

codyoss:
thanks!

EdSchouten:
Sorry, but I am going to close this PR. This is taking too long to get merged. I think I'll just hold on to this change locally.

PR #241, opened 2018-12-04
gomock: documents exported functions

### LINKED ISSUE #222: Document exported methods
A number of exported methods are undocumented, and therefore their behaviour can only be discovered by reading their code. For example, `Controller.Finish()` can only be called once, otherwise the test will fail, but this is not documented anywhere.

poy:
fixes #222

### DIFF
--- a/gomock/controller.go
+++ b/gomock/controller.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// GoMock - a mock framework for Go.
+// Package gomock is a mock framework for Go.
 //
 // Standard usage:
 //   (1) Define an interface that you wish to mock.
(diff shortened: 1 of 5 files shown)


gomock/controller.go
```diff
@@ -92,6 +113,8 @@ type Controller struct {
 	finished      bool
 }
 
+// NewController a new Controller. It is the preferred way to create a
```
balshetzer: Is this missing a word? NewController creates a new Controller.

poy: Ah, thanks. I think it should be:

> NewController returns a new Controller.

gomock/controller.go
```diff
@@ -200,6 +226,8 @@ func (ctrl *Controller) Call(receiver interface{}, method string, args ...interf
 	return rets
 }
 
+// Finish checks to see if all the methods that were expected to be called
+// were called. It should be invoked for each Controller.
```
RJPercival: It'd be good to document the requirement that it only be invoked exactly once.

PR #245, opened 2018-12-05
Test helper in mock

### LINKED ISSUE #239: mocks should call t.Helper()
The mocks are being flagged as a failing line during a test failure. This is due to the fact that `Controller.Call()`. invokes `t.Helper()`, but the mock does not. This means that instead of the offending line being called out, the mock is.

This could be solved by having the mock also call `t.Helper()`.

poy:
fixes #239

### DIFF
--- a/gomock/controller.go
+++ b/gomock/controller.go
@@ -70,45 +70,73 @@ type TestReporter interface {
 	Fatalf(format string, args ...interface{})
 }
 
+// TestHelper is a TestReporter that has the Helper method.  It is satisfied
+// by the standard library's *testing.T.
+type TestHelper interface {
+	TestReporter
+	Helper()
+}
+
 // A Controller represents the top-level control of a mock ecosystem.
 // It defines the scope and lifetime of mock objects, as well as their expectations.
 // It is safe to call Controller's methods from multiple goroutines.
 type Controller struct {
+	// T should only be called within a generated mock. It is not intended to
+	// be used in user code and may be changed in future versions. T is the
+	// TestReporter passed in when creating the Controller via NewController.
+	// If the TestReporter does not implment a TestHelper it will be wrapped
+	// with a nopTestHelper.
+	T             TestHelper
 	mu            sync.Mutex
-	t             TestReporter
 	expectedCalls *callSet
 	finished      bool
 }
 
 func NewController(t TestReporter) *Controller {
+	h, ok := t.(TestHelper)
+	if !ok {
+		h = nopTestHelper{t}
+	}
+
 	return &Controller{
-		t:             t,
+		T:             h,
 		expectedCalls: newCallSet(),
 	}
 }
 
 type cancelReporter struct {
-	t      TestReporter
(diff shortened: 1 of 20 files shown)


gomock/controller.go
```diff
@@ -70,62 +70,84 @@ type TestReporter interface {
...
+	Helper()
+}
+
 // A Controller represents the top-level control of a mock ecosystem.
 // It defines the scope and lifetime of mock objects, as well as their expectations.
 // It is safe to call Controller's methods from multiple goroutines.
 type Controller struct {
 	mu            sync.Mutex
-	t             TestReporter
+	T             TestHelper
```
balshetzer: I don't love that T is now a public var. I understand that it needs to be accessed by the mock. It's unfortunate that the testing package API requires direct access to T for T.Helper. I think we should at least add a comment that T is only public so that the mock can use it and any attempt by the user to set it, modify it, call it, etc. is unsupported and could be broken in the future without any consideration for backward compatibility (or something like that).

I wish T.Helper took a number like most other call stack related functions.

gomock/controller.go
```diff
@@ -70,62 +70,84 @@ type TestReporter interface {
...
-	r.t.Fatalf(format, args...)
+	r.TestHelper.Fatalf(format, args...)
 }
 
 // WithContext returns a new Controller and a Context, which is cancelled on any
 // fatal failure.
 func WithContext(ctx context.Context, t TestReporter) (*Controller, context.Context) {
+	h, ok := t.(TestHelper)
+	if !ok {
+		// TODO
```
balshetzer: Do now?

gomock/controller.go
poy: I will add a comment. I agree it is unfortunate, but I was unable to think of a way around it.

gomock/controller.go
poy: Ah, missed that one. Thanks.

poy:
@balshetzer PTAL

poy:
@balshetzer PTAL

gomock/controller.go
```diff
@@ -70,62 +70,88 @@ type TestReporter interface {
...
 	defer r.cancel()
-	r.t.Fatalf(format, args...)
+	r.TestHelper.Fatalf(format, args...)
 }
 
 // WithContext returns a new Controller and a Context, which is cancelled on any
 // fatal failure.
 func WithContext(ctx context.Context, t TestReporter) (*Controller, context.Context) {
+	h, ok := t.(TestHelper)
+	if ok {
```
balshetzer: if !ok {

poy:
@balshetzer PTAL

mikhailkolesnik:
can you pease tag this version. When we use `go get ` to install mockgen and `dep` to install `github.com/golang/mock` dep thinks that 1.1.1 is the most recent version and mockgen generates code that is not compatible with it.

poy:
@mikhailkolesnik tagged v1.2.0

PR #248, opened 2018-12-06
mockgen: package _test prepend package name

### LINKED ISSUE #247: mockgen: does not handle types if mocks are in test package
It looks like mockgen doesn't attach the package name to a type if the package is set (via `--package`).

Example:

Notice in `mock_test.go` any usage of `User` does not have the package name `users`.

**user.go**
```go
package users

type User struct {
	Name string
}

type Finder interface {
	FindUser(name string) User
}
```

```
mockgen --source=user.go --destination=mock_test.go --package=users_test
```

**mock_test**
```go
// Source: user.go

// Package users_test is a generated GoMock package.
package users_test

import (
	gomock "github.com/golang/mock/gomock"
	reflect "reflect"
)

// MockFinder is a mock of Finder interface
type MockFinder struct {
	ctrl     *gomock.Controller
	recorder *MockFinderMockRecorder
}

// MockFinderMockRecorder is the mock recorder for MockFinder
type MockFinderMockRecorder struct {
	mock *MockFinder
}

// NewMockFinder creates a new mock instance
func NewMockFinder(ctrl *gomock.Controller) *MockFinder {
	mock := &MockFinder{ctrl: ctrl}
	mock.recorder = &MockFinderMockRecorder{mock}
	return mock
}

// EXPECT returns an object that allows the caller to indicate expected use
func (m *MockFinder) EXPECT() *MockFinderMockRecorder {
	return m.recorder
}

// FindUser mocks base method
func (m *MockFinder) FindUser(name string) User {
	ret := m.ctrl.Call(m, "FindUser", name)
	ret0, _ := ret[0].(User)
	return ret0
}

// FindUser indicates an expected call of FindUser
func (mr *MockFinderMockRecorder) FindUser(name interface{}) *gomock.Call {
	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "FindUser", reflect.TypeOf((*MockFinder)(nil).FindUser), name)
}
```

poy:
fixes #247


mockgen/internal/tests/mock_in_test_package/README.md
```diff
@@ -0,0 +1,18 @@
+When a mock is put into the test package (e.g., `foo` vs `foo_test`), any type
```
balshetzer: This description reads well right now that the bug exists but in the future it will be confusing to understand. It would be better to write it from that perspective.

It should simply explain the situation it is testing. You can mention that this used to be broken but that will be of little interest the moment this is committed

mockgen/mockgen.go
```diff
@@ -223,6 +224,12 @@ func sanitize(s string) string {
 }
 
 func (g *generator) Generate(pkg *model.Package, pkgName string, outputPackagePath string) error {
+	// If the pkgName has _test as a suffix, then the package must be
+	// prepended.
+	if strings.HasSuffix(pkgName, "_test") {
```
balshetzer: Are you sure this is the right test? For example, the input package could be the test package. Wouldn't the right test be that the input and output packages are not the same packages?

mockgen/internal/tests/mock_in_test_package/README.md
poy: How about something far simpler:

> Test the case where the package has the `_test` suffix.

mockgen/mockgen.go
poy: That is what I thought originally, however it turns out to be a decent test as the case you described is already covered. For example if you look at the sample, it places the mocks in a new directory and the package is `mock_sample`. This all works.

The `test` package is a special case where it co-exists in the same directory, but has a different package name. Most of the mockgen code is looking at directories, and not package names.

balshetzer: Sorry about that. I think I wrote this part and assumed that each dir only has one package. Must have forgotten about the _test exception.

I can't really tell from the code if this is going to always do the right thing. Would you mind adding a test case where the input and out are the same package ending in _test? That's the case where I'm concerned this might not do the right thing.

poy: I added a test and found after further investigation that I could do as you originally suggested. I've switched to this method.

poy:
@balshetzer PTAL

PR #253, opened 2018-12-12
use go modules

poy:
Use golang.org/x/tools/go/packages instead of go/build as packages knows
how to deal with modules.


perkimon:
This would be really useful.

Issue #276, opened 2019-03-18
Mockgen generates wrong import path when vendor dir is directly under $GOPATH/src


ChuntaoLu:
When an imported package is inside `$GOPATH/src/vendor`, mockgen generates an incorrect import path that is prefixed with `vendor/`.

Suppose we have file `$GOPATH/src/vendor/foo/foo.go`
``` go
package foo

type Foo int
```
and `$GOPATH/src/bar/bar.go`
``` go
package bar

import "foo"

type Bar interface {
	Do() foo.Foo
}
```

Invoking command `mockgen bar Bar` gives the following output:
``` go
// Code generated by MockGen. DO NOT EDIT.
// Source: bar (interfaces: Bar)

// Package mock_bar is a generated GoMock package.
package mock_bar

import (
	gomock "github.com/golang/mock/gomock"
	reflect "reflect"
	foo "vendor/foo"
)

// MockBar is a mock of Bar interface
type MockBar struct {
	ctrl     *gomock.Controller
	recorder *MockBarMockRecorder
}

// MockBarMockRecorder is the mock recorder for MockBar
type MockBarMockRecorder struct {
	mock *MockBar
}

// NewMockBar creates a new mock instance
func NewMockBar(ctrl *gomock.Controller) *MockBar {
	mock := &MockBar{ctrl: ctrl}
	mock.recorder = &MockBarMockRecorder{mock}
	return mock
}

// EXPECT returns an object that allows the caller to indicate expected use
func (m *MockBar) EXPECT() *MockBarMockRecorder {
	return m.recorder
}

// Do mocks base method
func (m *MockBar) Do() foo.Foo {
	m.ctrl.T.Helper()
	ret := m.ctrl.Call(m, "Do")
	ret0, _ := ret[0].(foo.Foo)
	return ret0
}

// Do indicates an expected call of Do
func (mr *MockBarMockRecorder) Do() *gomock.Call {
	mr.mock.ctrl.T.Helper()
	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "Do", reflect.TypeOf((*MockBar)(nil).Do))
```

Notice the generated mock package has wrong import path for package foo ` foo "vendor/foo"`, it should have been `foo "foo"` instead.

PR #277, opened 2019-03-18
Fix import path for packages inside $GOPATH/src/vendor

ChuntaoLu:
This PR changes the search string of a vendor package's import path from `/vendor/` to  `vendor/` so that packages inside `$GOPATH/src/vendor` are properly imported in the generated mocks.

The reasoning is that `t.PkgPath()` for package `$GOPATH/src/vendor/foo` is `vendor/foo` while current code only checks against `/vendor/`, which does not cover the packages inside `$GOPATH/src/vendor`.

This PR fixes https://github.com/golang/mock/issues/276.

### DIFF
--- a/mockgen/model/model.go
+++ b/mockgen/model/model.go
@@ -359,15 +359,8 @@ func typeFromType(t reflect.Type) (Type, error) {
 	}
 
 	if imp := t.PkgPath(); imp != "" {
-		// PkgPath might return a path that includes "vendor"
-		// These paths do not compile, so we need to remove everything
-		// up to and including "/vendor/"
-		// see https://github.com/golang/go/issues/12019
-		if i := strings.LastIndex(imp, "/vendor/"); i != -1 {
-			imp = imp[i+len("/vendor/"):]
-		}
 		return &NamedType{
-			Package: imp,
+			Package: impPath(imp),
 			Type:    t.Name(),
 		}, nil
 	}
@@ -451,4 +444,18 @@ func typeFromType(t reflect.Type) (Type, error) {
 
 	// TODO: Struct, UnsafePointer
 	return nil, fmt.Errorf("can't yet turn %v (%v) into a model.Type", t, t.Kind())
+}
+
+// impPath sanitizes the package path returned by `PkgPath` method of a reflect Type so that
+// it is importable. PkgPath might return a path that includes "vendor". These paths do not
+// compile, so we need to remove everything up to and including "/vendor/".
+// See https://github.com/golang/go/issues/12019.
+func impPath(imp string) string {
+	if strings.HasPrefix(imp, "vendor/") {
+		imp = "/" + imp
+	}
+	if i := strings.LastIndex(imp, "/vendor/"); i != -1 {
+		imp = imp[i+len("/vendor/"):]
+	}
+	return imp
 }
--- a/mockgen/model/model_test.go
+++ b/mockgen/model/model_test.go
@@ -17,9 +17,14 @@ func TestImpPath(t *testing.T) {
 	}{
 		{"foo/bar", "foo/bar"},
 		{"vendor/foo/bar", "foo/bar"},
+		{"vendor/foo/vendor/bar", "bar"},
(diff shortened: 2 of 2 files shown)


ChuntaoLu:
I signed it!

ChuntaoLu:
@poy Could you take a look at this PR? Thank you.

balshetzer:
It seems like this change will make us treat any directory that ends with "vendor" as if it were a vendor directory. For example, a directory called "rivendor" would be mistakenly treated as a "vendor" directory. Do you think you could fix that? Ideally, you would also add a test.

ChuntaoLu:
> It seems like this change will make us treat any directory that ends with "vendor" as if it were a vendor directory. For example, a directory called "rivendor" would be mistakenly treated as a "vendor" directory. Do you think you could fix that? Ideally, you would also add a test.

Thanks for taking a look. That is a good catch, I will fix that and add a test.

ChuntaoLu:
@balshetzer I actually don't know how to do a proper test with the current CI setup. I tried to use a different [GOPATH](https://github.com/golang/mock/pull/277/files#diff-f0c23b53df111ece0b41e2e78e6ee32f) to simulate the case where vendor is under `$GOPATH/src`, but that does not play well with the CI job. Any suggestions?

balshetzer:
You could try just unit testing a function that does the job.

ChuntaoLu:
@balshetzer Added unit tests for the fix, please take a another look. Thank you.

mockgen/model/model.go
```diff
@@ -452,3 +445,17 @@ func typeFromType(t reflect.Type) (Type, error) {
...
 	return nil, fmt.Errorf("can't yet turn %v (%v) into a model.Type", t, t.Kind())
 }
+
+// impPath sanitizes the package path returned by `PkgPath` method of a reflect Type so that
+// it is importable. PkgPath might return a path that includes "vendor". These paths do not
+// compile, so we need to remove everything up to and including "/vendor/".
+// See https://github.com/golang/go/issues/12019.
+func impPath(imp string) string {
+	if strings.HasPrefix(imp, "vendor/") {
+		imp = "/" + imp
```
balshetzer: return imp[7:]

mockgen/model/model_test.go
```diff
@@ -0,0 +1,32 @@
...
+		{"qux/vendor/foo/vendor/bar", "bar"},
+	}
+	for _, tc := range testCases {
+		t.Run(fmt.Sprintf("input %s", tc.input), func(t *testing.T) {
+			if got := impPath(tc.input); got != tc.want {
+				t.Errorf("got %s; want %s", got, tc.want)
+			}
+		})
+	}
+
```
balshetzer: remove blank line

mockgen/model/model_test.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+	testCases := []struct {
+		input string
+		want  string
+	}{
+		{"foo/bar", "foo/bar"},
+		{"vendor/foo/bar", "foo/bar"},
+		{"/vendor/foo/bar", "foo/bar"},
+		{"qux/vendor/foo/bar", "foo/bar"},
+		{"qux/vendor/foo/vendor/bar", "bar"},
```
balshetzer: Add some test cases for the bug we found in the last review, where the word vendor appears as part of the directory:
"govendor/foo"
"foo/govendor/bar"
"vendors/foo"
"foo/vendors/bar"

mockgen/model/model.go
ChuntaoLu: Is it possible that the path can be something like `vendor/foo/vendor/bar`? If that's possible, then we need to return `bar` instead of `foo/vendor/bar`, right?

mockgen/model/model_test.go
ChuntaoLu: done

mockgen/model/model_test.go
ChuntaoLu: done. also added one for `vendor/foo/vendor/bar`

PR #292, opened 2019-05-10
fix callserrors with empty msg

### LINKED ISSUE #291: Empty Error message in Concurrent mode
This is a weird bug in concurrent mode, gomock will fatal and print empty error msg, it's really confused me

Detail description:

My application has a producer(in one step) and consumer (in several step).They are running in seperated goroutines
```
type producer interface{
       pull()
}

type consumer interface{
      getMessage()
      consume()
      replyMessage()
}
```

and I use mock to write 
```
gomock.InOrder(
     pull().anyTimes(),
    getMessage().anyTimes(),
    consume().anyTimes(),
    replyMessage().anyTimes(),
)

```

I know it's not ok to write it in order, correct me later pls.

And one of the actual invoke order is 
```
pull()
getMessge()
pull()
consume()
replyMessage()
```
the second pull will be detected by gomock,but it only print empty error msg for this method

I listed the real fatal message it printed below
```
=== RUN   TestFlow_RunForever
wait for 3 second
pull msg from cmq
download from cos
build and sleep 10s
received signal terminated and wait for terminating
1.close message channel
2.wait for task done
push and sleep 5s
create image in db
send msg back to cmq
delete msg from cmq
3.terminated
--- FAIL: TestFlow_RunForever (20.00s)
    mq.go:45: Unexpected call to *cmqmock.MockProcessor.Pull([0xc0000b8b90]) at /Users/{user}/Code/golang/{myapp}/mock/cmq/mock.go:54 because: 
FAIL

```
=================

After debuging  it ,i found the reason is hiden in gomock/callset.go:66.

```
// FindMatch searches for a matching call. Returns error with explanation message if no call matched.
func (cs callSet) FindMatch(receiver interface{}, method string, args []interface{}) (*Call, error) {
	key := callSetKey{receiver, method}

	// Search through the expected calls.
	expected := cs.expected[key]
	var callsErrors bytes.Buffer
	for _, call := range expected {
		err := call.matches(args)
		if err != nil {
			fmt.Fprintf(&callsErrors, "\n%v", err)
		} else {
			return call, nil
		}
	}

	// If we haven't found a match then search through the exhausted calls so we
	// get useful error messages.
	exhausted := cs.exhausted[key]
	for _, call := range exhausted {
		if err := call.matches(args); err != nil {
			fmt.Fprintf(&callsErrors, "\n%v", err)
		}
	}

	if len(expected)+len(exhausted) == 0 {
		fmt.Fprintf(&callsErrors, "there are no expected calls of the method %q for that receiver", method)
	}

	return nil, fmt.Errorf(callsErrors.String())
}
```
The second pull will match empty expected and one exhausted,and ` call.matches(args)` in 
```
        for _, call := range exhausted {
		if err := call.matches(args); err != nil {
			fmt.Fprintf(&callsErrors, "\n%v", err)
		}
	}
```
will pass without error

So the flow go through everything and return with empty msg

mjaow:
Fix #291 

check exhaustedNum and callsErrors.len() , if it's true, fill with error msg

### DIFF
--- a/gomock/callset.go
+++ b/gomock/callset.go
@@ -87,8 +87,13 @@ func (cs callSet) FindMatch(receiver interface{}, method string, args []interfac
 		}
 	}
 
-	if len(expected)+len(exhausted) == 0 {
+	expectedNum := len(expected)
+	exhaustedNum := len(exhausted)
+
+	if expectedNum+exhaustedNum == 0 {
 		fmt.Fprintf(&callsErrors, "there are no expected calls of the method %q for that receiver", method)
+	} else if exhaustedNum > 0 && callsErrors.Len() == 0 {
+		fmt.Fprintf(&callsErrors, "there are no expected calls of the method %q in that order", method)
 	}
 
 	return nil, fmt.Errorf(callsErrors.String())
--- a/gomock/callset.go
+++ b/gomock/callset.go
@@ -90,7 +90,7 @@ func (cs callSet) FindMatch(receiver interface{}, method string, args []interfac
 	expectedNum := len(expected)
 	exhaustedNum := len(exhausted)
 
-	if expectedNum+exhaustedNum == 0 {
+	if expectedNum == 0 && exhaustedNum == 0 {
 		fmt.Fprintf(&callsErrors, "there are no expected calls of the method %q for that receiver", method)
 	} else if exhaustedNum > 0 && callsErrors.Len() == 0 {
 		fmt.Fprintf(&callsErrors, "there are no expected calls of the method %q in that order", method)


balshetzer:
I don't understand this fix. Can you explain it? 

If the reason a call doesn't match an expectation is an InOrder that matches will return an error saying that. So why do you believe that `f exhaustedNum > 0 && callsErrors.Len() == 0` indicates an ordering issue?

mjaow:
@balshetzer 
i thought it was a order issue caused in concurrent invoke.
1.callsErrors.Len() means all method matches
2.exhaustedNum >0 (and expectedNum==0) means the method shouldn't be here, should be in prev position
according two reason,I concluded it was a order issue

Sorry,I couldn't think of another better reason to describe and I thought it was the simplest way to fix it

cvgw:
Some additional info
There is a case where `matches` will return nil for an ordering error.
e.g.
```
gomock.InOrder(
  mockFoo.EXPECT().Foo().AnyTimes(),
  mockBar.EXPECT().Bar().AnyTimes(),
  mockBaz.EXPECT().Baz().AnyTimes(),
)

mockFoo.Foo()
mockBar.Bar()
mockFoo.Foo()
```
This will produce an error message with no additional information following `because: ` (as reported in the issue)

`err` will always be nil https://github.com/golang/mock/blob/1bf84fa8d4e5fedaedb9b103bb5b947e25bf5d8e/gomock/callset.go#L85

`len(exhausted)` will be 1 and `len(expected)` will be 0
https://github.com/golang/mock/blob/1bf84fa8d4e5fedaedb9b103bb5b947e25bf5d8e/gomock/callset.go#L90

gomock/callset.go
```diff
@@ -87,8 +87,13 @@ func (cs callSet) FindMatch(receiver interface{}, method string, args []interfac
 		}
 	}
 
-	if len(expected)+len(exhausted) == 0 {
+	expectedNum := len(expected)
+	exhaustedNum := len(exhausted)
+
+	if expectedNum == 0 && exhaustedNum == 0 {
 		fmt.Fprintf(&callsErrors, "there are no expected calls of the method %q for that receiver", method)
+	} else if exhaustedNum > 0 && callsErrors.Len() == 0 {
```
cvgw: Is `exhaustedNum` required? At this point of the function we know we are in error behavior. Are there any other cases where `callsErrors.Len() == 0`?

cvgw: I think a comment here explaining what error scenario we are handling (i.e. an ordering error) and the rational behind the check for this scenario would be really good.

gomock/callset.go
```diff
@@ -87,8 +87,13 @@ func (cs callSet) FindMatch(receiver interface{}, method string, args []interfac
 		}
 	}
 
-	if len(expected)+len(exhausted) == 0 {
+	expectedNum := len(expected)
+	exhaustedNum := len(exhausted)
+
+	if expectedNum == 0 && exhaustedNum == 0 {
```
cvgw: Is this check safe to change? I don't think it is possible for len to ever be less than zero so it seems fine to me.

cvgw:
@mjaow are you interested in continuing to work on this PR? If not I'm happy to take it over.

codyoss:
Due to lack of response I am closing out this PR for now. I created #404 to track this issue, assigning it to @cvgw for now. Thank you for reporting this.

PR #334, opened 2019-10-18
Update documentation on MinTimes and MaxTimes. Closes #331

### LINKED ISSUE #331: Calling EXPECT().MinTimes(1).MaxTimes(1) / MaxTimes(1).MinTimes(1) is confusing
The documentation says:

>  If AnyTimes or MaxTimes have not been called, MinTimes sets the maximum number of calls to infinity.

But if MaxTimes is called with argument 1 the MinTimes code considers this as MaxTimes has not been called yet.
See 
https://github.com/golang/mock/blob/dd8d2a22370e4c8a334e80ca8477f71356c8e4bb/gomock/call.go#L89-L91
and
https://github.com/golang/mock/blob/dd8d2a22370e4c8a334e80ca8477f71356c8e4bb/gomock/call.go#L99-L101

This is confusing and should either be fixed in the documentation or by adding a flag to check whether Min/Max-Times was called.

rkpatel7:
Updated the documentation to something I think seems to be clearer. Also added two tests for this.

### DIFF
--- a/gomock/call.go
+++ b/gomock/call.go
@@ -82,8 +82,8 @@ func (c *Call) AnyTimes() *Call {
 	return c
 }
 
-// MinTimes requires the call to occur at least n times. If AnyTimes or MaxTimes have not been called, MinTimes also
-// sets the maximum number of calls to infinity.
+// MinTimes requires the call to occur at least n times. If AnyTimes or MaxTimes have not been called or if MaxTimes
+// was previously called with 1, MinTimes also sets the maximum number of calls to infinity.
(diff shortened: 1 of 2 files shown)


rkpatel7:
It's a little unfortunate the CLA requires the commit to include a public email address. It would be nice if it worked with the github no-reply email address. I ended up having to make a separate gmail account to sign the CLA and associate with this commit.

codyoss:
@rkpatel7 Yeah, I am not sure if there is a way around that one. But I appreciate you going thought all of the trouble!

Issue #190, opened 2018-05-23
Error output is ambiguous for equivalent values with different types


bradwilzon:
https://github.com/golang/mock/blob/22bbf0ddf08105dfa364d0a2fa619dfa71014af5/gomock/matchers.go#L53

When printing output for matched errors, if the expected and actual value have different types, either a special message should be printed (ie. expected value type does not match actual value type), or simply print the type with the output.

My confusing error output:
```
Expected call at mytest.go:467 doesn't match the argument at index 1.
Got: 8
Want: is equal to 8
```
In this case, I was using a `const val = 8` which is resolved to an `int`. The const is used in the innards of a struct that expects an `int32` (so the const is correctly casted). The const is then passed as the parameter to a method mocked by `EXPECT()` that originally takes an `int32`, but the mock version of the method takes an `interface{}`. Therefore, in the second case, the const is never casted correctly to int32 and causes a type mismatch further down the call path.

The long story short is, it's possible for gomock to print equivalent values without type information.

codyoss:
I agree that this seems like something we should do. I would be willing to look at PR for this. Types might add a bit of noise if always there. Could hide that info behind a verbose flag as well... If you liked this issue feel free to share your thoughts.

codyoss:
Another usecase from #267 
>When comparing an expected arg with a nil slice to an actual arg with a zero length slice, mock concludes the two do not match, however, the output for 'want' and 'got' is exactly the same for the nil/emtpy slice and does not indicate where the mismatch occurred.

Example:

--- FAIL: TestNilVsEmpty (0.00s)
    example_test.go:24: Unexpected call to *mock_example.MockThing.Apply([0xc0000900a0]) at /usr/local/google/home/briankennedy/golang/work/src/github.com/briantkennedy/mockbug/pkg/mock_example/mock_thing.go:39 because: 
        Expected call at /usr/local/google/home/briankennedy/golang/work/src/github.com/briantkennedy/mockbug/pkg/example/example_test.go:19 doesn't match the argument at index 0.
        Got: &{[]}
        Want: is equal to &{[]}
    asm_amd64.s:522: missing call(s) to *mock_example.MockThing.Apply(is equal to &{[]}) /usr/local/google/home/briankennedy/golang/work/src/github.com/briantkennedy/mockbug/pkg/example/example_test.go:19
    asm_amd64.s:522: aborting test due to missing call(s)
FAIL
FAIL	github.com/briantkennedy/mockbug/pkg/example	0.021s
[...]

PR #341, opened 2019-11-03
Make error message clearer

adzimzf:
Referring to [this](https://github.com/golang/mock/issues/190) issue.
Add data type on error message if the data type is not equal

### DIFF
--- a/gomock/call.go
+++ b/gomock/call.go
@@ -301,8 +301,8 @@ func (c *Call) matches(args []interface{}) error {
 
 		for i, m := range c.args {
 			if !m.Matches(args[i]) {
-				return fmt.Errorf("Expected call at %s doesn't match the argument at index %s.\nGot: %v\nWant: %v",
-					c.origin, strconv.Itoa(i), args[i], m)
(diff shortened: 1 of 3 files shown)


adzimzf:
@googlebot I signed it!

codyoss:
Thanks for your contribution, I like the idea of this change. I am just thinking wether something like this should be put behind a verbose flag... Thinking...

codyoss:
@cvgw @nguyenfilip What are your thoughts on this? Should something like this be default behavior, or should we hide types behind the verbose flag?!?

cvgw:
@codyoss my initial thought is that I would want the type info returned when needed without having to enable a flag. What would "when needed" mean? Perhaps an initial check whether the types are equal before evaluating further?  That way an error message specific to the type could be generated and the rest of the error messages could exclude type information.

Looking at the current code I think that would be a non-trivial effort and would require some refactoring of where error messages are generated.

codyoss:
I like that idea. Also I agree, it would be a non-trivial effort.

codyoss:
I am going to close this for now as I think we need a more robust solution. I think this is still an issue and something we need to work towards.

Issue #342, opened 2019-11-05
-self_package does not work when -package is different from the base dir of package path


linzhp:
# Steps to reproduce:

1. Clone git@github.com:linzhp/go_examples.git to GOPATH
2. Run `mockgen -package core -self_package github.com/linzhp/go_examples/interfaces github.com/linzhp/go_examples/interfaces Methods`

# What do I expect:
The generated mock should not import `github.com/linzhp/go_examples/interfaces` because it is specified in `-self_package`

# What do I see instead:
```go
import (
	gomock "github.com/golang/mock/gomock"
	interfaces "github.com/linzhp/go_examples/interfaces"
	reflect "reflect"
)
```

codyoss:
Thanks for the report, I will check out your PR soon here!

codyoss:
I don't think there is an issue here, or at least not with the example you have given. In this example your interface returns a struct, `Info`, from the `core` package. Because of this, the mock must include `	interfaces "github.com/linzhp/go_examples/interfaces"`

The output that makes this required:

```go
// getInfo mocks base method
func (m *MockMethods) getInfo() interfaces.Info {
	m.ctrl.T.Helper()
	ret := m.ctrl.Call(m, "getInfo")
	ret0, _ := ret[0].(interfaces.Info)
	return ret0
}
```

Thoughts?

linzhp:
> Because of this, the mock must include interfaces "github.com/linzhp/go_examples/interfaces"

This is only true when the mock is going to be compiled in a different package. However, the intention here, by specifying `-self_package github.com/linzhp/go_examples/interfaces`, is to generate the mock and compile it as part of the package `github.com/linzhp/go_examples/interfaces`. In other words, the mock and `Info` will be in the same package, so I don't want the import `github.com/linzhp/go_examples/interfaces`.

Is this a valid use of the `-self_package` argument?

codyoss:
Thanks for the clarification. I did not realize the import path did not match package name

For now as a workaround this should work:
`mockgen -package core -self_package github.com/linzhp/go_examples/interfaces -source=types.go`

Still thinking about the change, just want to make sure if does not break other things.

linzhp:
Yes, source mode works as a workaround. However, source mode is slower under [bazel_gomock](https://github.com/jmhodges/bazel_gomock) rule, as it requires constructing a GOPATH in the sandbox.

Thanks for looking into this.

codyoss:
Fixed with #343.

PR #343, opened 2019-11-06
Always remove self_package import

linzhp:
When users provide `-self_package`, gomock should always exclude that import from generated mock. The `pkg.Name` can be incorrectly determined in reflect mode:
 https://github.com/golang/mock/blob/0b73a1dbc3e1197079e9b9db31ab2879a5c9a095/mockgen/reflect.go#L206-L211

We should not reset `outputPackagePath` when it is passed from `-self_package`.

Fixing #342


codyoss:
Not sure why this is working on CI, but if you run `go install ./... && ./ci/check_go_generate.sh` locally this fails.

codyoss:
This may having nothing to do with this PR btw, but would like to get it fixed before evaluating further PRs.

codyoss:
@linzhp thanks for your patience. I wanted to make sure we got our build properly working on master before we accepted any more PRs. Would you mind rebasing this as there are now some conflicts.

linzhp:
@codyoss rebased

linzhp:
Are you planning another release soon? It's been 8 months

codyoss:
Yes, originally planned for tomorrow: https://github.com/golang/mock/milestone/1
Should land sometime this week. This package recently switched maintainers, so there was a bit of a lull.

codyoss:
@linzhp One more thing, would you mind adding a testcase under `mockgen/internal/tests`. All it would need to be is a source file with a `go: generate` directive and the generated code. Our ci will try to regenerate all these and check for diffs to make sure this does not break again in the future.

linzhp:
@codyoss Added a test case

codyoss:
@linzhp Thank you for your patience and contribution! Should have this rolled up in a release later this week 🎆

Issue #250, opened 2018-12-10
Support --version flag or similar option in mockgen


a1gemmel:
It would be nice to be able to know what version of mockgen I have installed via a flag or command, like:

`mockgen -v`
`mockgen -version`
`mockgen --version`
`mockgen version`

Any of the above, really.

poy:
This is a great idea. The only hard part is how to supply the version. I think most developers pull `mockgen` via `go get`, and not from a github release. Therefore we can't necessarily use compile flags or something like that to supply the version to the binary. This implies we have to have a constant in the code capturing the current version (which is a little weird).

I think if we want to supply this, then we need to start publishing this stuff to github and encourage people to get it from there.

@balshetzer Do you have any thoughts?

Rossiar:
@codyoss @poy any plans to work on this soon? I've just spent half a day debugging a `-self_package` issue and it was all because I was pulling master and not the latest release (1.3.1)

Right now on master the `README.md` tells users to run:
```
go get github.com/golang/mock/mockgen
go install github.com/golang/mock/mockgen
``` 
These commands will just install the latest code from `master` - which right now seems to come with a bunch of unwanted side-effects (as I've found out today).

I think using compile flags and uploading the binaries with the release is a really good approach going forward - adding the `--version` option as well I think will clear up a lot of confusion when reporting issues - our team really loves using `gomock` so I'd be happy to offer some of our time to implement this feature (with your guidance).

nguyenfilip:
@Rossiar I can look a bit more into this. @codyoss let me know in case you already have some ideas how to implement this. Otherwise I will do some research so that both `go get` and direct binary usage works reasonably.

codyoss:
I know some libraries just have a constant string that gets exposed. We could write a release script that automatically bumps it, etc. I have no strong opinions.

nguyenfilip:
My suggestion of how to address this is in  #362 . Since most of the people use `go get` I think we can continue supporting that workflow. Version information is available if _module aware mode_ is used. In GOPATH mode I suggest we print a warning message for the users informing them they are running some random version of the binary.

I think exposed constant string is a little bit of a hassle. Lets say we bump the constant string to `v1.4.0` in commit X and then over the course of a few days we add a few more commits. Anybody who runs `go get` since X will see `gomock -version` to output `v1.4.0`, regardless of the commits added on top of X.

BradErz:
to be perfectly honest the way we implement this for our services is 

`Makefile`
```Makefile
[...]

PR #362, opened 2019-12-20
Add Mockgen Version Flag

nguyenfilip:
Add -version flag that shows a version provided by Go modules.

This strategy works well for module aware invocations of `go get` and `go
run` (which means using GO111MODULE=on).

If we ever plan to distribute built binaries, it is necessary to amend
the strategy with build flags.

GOPATH mode versioning is unsupported.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,14 +8,17 @@ contexts too.
 Installation
 ------------
 
-Once you have [installed Go][golang-install], install the `mockgen` tool:
+Once you have [installed Go][golang-install], install the `mockgen` tool.
+
+To get the latest released version use:
 
 ```bash
-go get github.com/golang/mock/mockgen
+GO111MODULE=on go get github.com/golang/mock/mockgen@latest
 ```
 
-_Note: It is recommended to have `GO111MODULE=on` to ensure the correct
-dependencies are used._
+If you use `mockgen` in your CI pipeline, it may be more appropriate to fixate
+on a specific mockgen version.
+
 
 Documentation
 -------------
--- a/mockgen/version.1.11.go
+++ b/mockgen/version.1.11.go
@@ -1,4 +1,4 @@
-// Copyright 2010 Google Inc.
+// Copyright 2019 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/mockgen/version.1.11.go
+++ b/mockgen/version.1.11.go
@@ -1,0 +1,26 @@
+// Copyright 2010 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build !go1.12
+
+package main
+
+import (
+	"log"
+)
+
+func printVersion() {
+	log.Printf("No version information is available for Mockgen compiled with " +
+		"version 1.11")
+}
--- a/mockgen/version.1.12.go
+++ b/mockgen/version.1.12.go
@@ -1,4 +1,4 @@
-// Copyright 2010 Google Inc.
+// Copyright 2019 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/mockgen/version.1.12.go
+++ b/mockgen/version.1.12.go
@@ -1,0 +1,35 @@
+// Copyright 2010 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+// +build go1.12
+
+package main
+
+import (
+	"fmt"
+	"log"
+	"runtime/debug"
+)
+
+func printVersion() {
+	if bi, exists := debug.ReadBuildInfo(); exists {
+		fmt.Println(bi.Main.Version)
+	} else {
+		log.Printf("No version information found. Make sure to use " +
+			"GO111MODULE=on when running 'go get' in order to use specific " +
+			"version of the binary.")
+	}
+
+}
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -8,14 +8,17 @@ contexts too.
...
 ------------
 
-Once you have [installed Go][golang-install], install the `mockgen` tool:
+Once you have [installed Go][golang-install], install the `mockgen` tool.
+
+To get the latest released version use:
 
 ```bash
-go get github.com/golang/mock/mockgen
+GO111MODULE=on go get github.com/golang/mock/mockgen@latest
```
codyoss: These extra options should not be needed with the current version of go right?

README.md
```diff
@@ -8,14 +8,17 @@ contexts too.
...
 ```
 
-_Note: It is recommended to have `GO111MODULE=on` to ensure the correct
-dependencies are used._
+If you use `mockgen` in your CI pipeline, it may be more appropriate to fixate
```
codyoss: Eventually in a wiki we might want to show the tool install pattern used with modules. I can point to an example later.

mockgen/version.1.11.go
```diff
@@ -0,0 +1,26 @@
+// Copyright 2010 Google Inc.
```
codyoss:
```suggestion
// Copyright 2019 Google LLC
```
The guidance in LLC now.

mockgen/version.1.12.go
```diff
@@ -0,0 +1,35 @@
+// Copyright 2010 Google Inc.
```
codyoss:
```suggestion
// Copyright 2019 Google LLC
```

mockgen/version.1.12.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"fmt"
+	"log"
+	"runtime/debug"
+)
+
+func printVersion() {
+	if bi, exists := debug.ReadBuildInfo(); exists {
+		fmt.Println(bi.Main.Version)
```
cvgw: I haven't seen this pattern before (usually see a const str). Does this return the version metadata of the go module? Does that mean it won't work unless installed via module support? I'm guessing that's what the build checks and error messaging are about.

codyoss: What does this actually print out today if you run it on go1.12+?

README.md
nguyenfilip: Actually it is needed because `GO111MODULE` is by default set to `auto`. That means that if you invoke `go get` in a directory outside the `GOPATH` and without go.mod file, then the `go get` executes in  `GOPATH mode`. These are the docs:
```
 If GO111MODULE=auto or is unset, then the go command enables or disables module support based on the current directory. Module support is enabled only when the current directory contains a go.mod file or is below a directory containing a go.mod file.
```
(source: https://golang.org/cmd/go/#hdr-Module_support)

README.md
nguyenfilip: ok

mockgen/version.1.12.go
nguyenfilip: I am not leaning strongly towards the implementation in this PR, but I tend to think it's the best out of the three options that I have considered:

1. using constant string
2. using build flags
3. using debug.ReadBuildInfo

Each option has pros and cons. 

Problem with constant string is that any time somebody pulls master version X (not the latest version) of mockgen with `go get`, they will always end up with mockgen that includes all the commits since X but the version string is unchanged (the binary would report same version for binaries with different commits, which is exactly what we dont want).

Problem with build flags is that they won't work unless we distribute binaries. And I see that current assumption is that people use `go get` to get the mockgen

Problem with debug.ReadBuildInfo is that it relies on using modules. 

I lean towards using debug.ReadBuildInfo. I think it should be the best practice for the users of `mockgen` to always use predictable version, especially for CI builds. That was the original problem in #250. Thus I think that users should not run `go get` in GOPATH mode and they should use `@` to specify which version of mockgen they want to use. Note that the version info is embedded in the binary, so after `go get` is used to download the binary the clients can copy the `mockgen` out of the GOPATH and `-version` should still work.

As I mentioned in the description of this PR (`the GOPATH mode versioning is unsupported`) this wont work unless installed via module support 

WDYT?

nguyenfilip: As soon as we release 1.3.2, the `-version` flag should report `v1.3.2` if `go get` is invoked with `@latest`. I think there is no way to test this with `mockgen` itself now. But I tested the `debug.ReadBuildInfo` in my other test project. 

One edge case is when somebody invokes the `go get` with `@master`. But it does behave correctly. With `@master` the resulting version should be something like: `v1.3.2-0.20191220092011-827cab0badc9`, in other words if the `go get` is requsted to download an unreleased version, the module support adds commit SHA to the version string to disambiguate the version.

There are multiple situations when `-version` is not going to be available (but I think that is fine):
 - when building/running the mockgen module locally the `-version` is `(devel)`
 - when running with Golang older then 1.12 (as seen in the changes in this PR)

cvgw: >Problem with constant string is that any time somebody pulls master version X (not the latest version) of mockgen with go get, they will always end up with mockgen that includes all the commits since X but the version string is unchanged (the binary would report same version for binaries with different commits, which is exactly what we dont want).

I've always used a constant string in projects but I haven't thought about the UX of pulling master in that case. I don't think I have a strong opinion and I don't disagree with the point you've made.

I don't think there is anything wrong with making module support a requirement for the version command to work; as long as we fall back to some other behavior when module support is not enabled.

cvgw: It looks like there is some nice fallback behavior using the build tags for different versions of go :👍

mockgen/version.1.11.go
```diff
@@ -0,0 +1,26 @@
...
+
+package main
+
+import (
+	"log"
+)
+
+func printVersion() {
+	log.Printf("No version information is available for Mockgen compiled with " +
+		"version 1.11")
```
codyoss: nit: < 1.12

README.md
codyoss: As a part of our next release I think we should start uploading artifacts as well. This will allow more people to get the desired behavior out of version. We can use a recent contribution to help with this: https://github.com/golang/mock/pull/229

mockgen/version.1.12.go
nguyenfilip: @cvgw wrt build tags. Are you referring to the `// +build go1.12` tag that has been in this PR?

mockgen/version.1.11.go
nguyenfilip: done, thanks!

README.md
nguyenfilip: Cool! Yes if we build binaries ourselves and we build it with 1.12+ version of golang, then the built binaries should support the -version flag as implemented here.

mockgen/version.1.12.go
nguyenfilip: > I've always used a constant string in projects but I haven't thought about the UX of pulling master in that case. I don't think I have a strong opinion and I don't disagree with the point you've made.

Yeah I agree. Constant string is a good solution if you build and distribute the binary. But! :-) If we did build and distribute the binary, it think there is even better solution than version string. The better solution is to use `-ldflags` with `go build` command. That solution allows you to set a value to global property in your golang program during build time.

cvgw: > @cvgw wrt build tags. Are you referring to the `// +build go1.12` tag that has been in this PR?

yes

>The better solution is to use `-ldflags` with `go build` command. That solution allows you to set a value to global property in your golang program during build time.

I realized that's what we do during compile on another project I help on https://github.com/GoogleContainerTools/kaniko/blob/master/Makefile#L30

PR #363, opened 2019-12-20
Update issue templates

codyoss:


codyoss:
I wonder if something with the heading was messing this up? Let's try and see.

cvgw:
That worked. Did you do that through the UI or did updating the file fix it?

codyoss:
I did purpose the change from the UI, but my hunch is this is related to the header info block 🤷‍♂

PR #368, opened 2019-12-20
add Len matcher

codyoss:
This matcher will check the length with any type that allows this
behavior via reflection. Implementation inspired by #189.

### DIFF
--- a/gomock/matchers.go
+++ b/gomock/matchers.go
@@ -176,6 +176,24 @@ func (am allMatcher) String() string {
 	return strings.Join(ss, "; ")
 }
 
+type lenMatcher struct {
+	i int
+}
+
+func (m lenMatcher) Matches(x interface{}) bool {
+	v := reflect.ValueOf(x)
+	switch v.Kind() {
+	case reflect.Array, reflect.Chan, reflect.Map, reflect.Slice, reflect.String:
+		return v.Len() == m.i
+	default:
+		return false
+	}
+}
+
+func (m lenMatcher) String() string {
+	return fmt.Sprintf("has length %d", m.i)
+}
+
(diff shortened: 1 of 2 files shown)


gomock/matchers.go
```diff
@@ -176,6 +176,24 @@ func (am allMatcher) String() string {
...
 }
 
+type lenMatcher struct {
+	i int
+}
+
+func (m lenMatcher) Matches(x interface{}) bool {
+	v := reflect.ValueOf(x)
+	switch v.Kind() {
+	case reflect.Array, reflect.Chan, reflect.Map, reflect.Slice, reflect.String:
```
cvgw: curious, does this cover `*[]string` and `[]string` ?

gomock/matchers.go
```diff
@@ -192,6 +210,12 @@ func Any() Matcher { return anyMatcher{} }
 //   Eq(5).Matches(4) // returns false
 func Eq(x interface{}) Matcher { return eqMatcher{x} }
 
+// Len returns a matcher that matches on length. This matcher returns false if
+// is compared to a type that is not an array, chan, map, slice, or string.
+func Len(i int) Matcher {
```
cvgw: nit: I'm imagining `mockFoo.EXPECT().Bar(gomock.Len(4))` which feels like it could be more natural but I can't think of a better alternative.

gomock/matchers.go
codyoss: This would not, do you think we should support that?

cvgw: I don't think it's necessarily needed. If it supports `[]string` but not `*[]string` the user can just dereference.

PR #371, opened 2019-12-27
fix generates non-compilable code

### LINKED ISSUE #370: Generates non-compilable code
I only need to document one specific parameter in the interface, an example:
```
Search(_ Ctx, _ Auth, query string,_ Lang,_ Page)
```

**Actual behavior** Generates non-compilable code, sample:

```
(m *MockSearcher) TrackByIDs(_ Ctx, _ UserID, subsUserIDs []UserID, _ []ulid.ULID) ([]TrackInfo, error) {
	m.ctrl.T.Helper()
	ret := m.ctrl.Call(m, "TrackByIDs", _, _, subsUserIDs, _)
	ret0, _ := ret[0].([]TrackInfo)
	ret1, _ := ret[1].(error)
	return ret0, ret1
}
```

**Expected behavior** I expected to see that result.
```
(m *MockSearcher) TrackByIDs(arg1 Ctx, arg2 UserID, subsUserIDs []UserID, arg4 []ulid.ULID) ([]TrackInfo, error) {
	m.ctrl.T.Helper()
	ret := m.ctrl.Call(m, "TrackByIDs", _, _, subsUserIDs, _)
	ret0, _ := ret[0].([]TrackInfo)
	ret1, _ := ret[1].(error)
	return ret0, ret1
}
```

1.  Write interface:
```
type Sample interface {
	Test(val int, _ string) 
}
```
2.  Generate mock.

**Additional Information**

-   gomock version or git ref: 1.3.1 
-   golang version: 1.13

ZergsLaw:
Fixes #370 .

**Description**
- [x] fixed bug

**Submitter Checklist**

- [x] Includes tests

### DIFF
--- a/mockgen/mockgen_test.go
+++ b/mockgen/mockgen_test.go
@@ -267,3 +267,89 @@ func findMethod(t *testing.T, identifier, methodName string, lines []string) int
 	t.Fatalf("unable to find 'func (m %s) %s'", identifier, methodName)
 	panic("unreachable")
 }
+
+func TestGetArgNames(t *testing.T) {
+	for _, testCase := range []struct {
+		name     string
+		method   *model.Method
+		expected []string
+	}{
+		{
+			name: "NamedArg",
+			method: &model.Method{
+				In: []*model.Parameter{
+					{
+						Name: "firstArg",
+						Type: &model.NamedType{Type: "int"},
+					},
+					{
+						Name: "secondArg",
+						Type: &model.NamedType{Type: "string"},
+					},
+				},
+			},
+			expected: []string{"firstArg", "secondArg"},
+		},
+		{
+			name: "NotNamedArg",
+			method: &model.Method{
+				In: []*model.Parameter{
+					{
+						Name: "",
+						Type: &model.NamedType{Type: "int"},
+					},
+					{
+						Name: "",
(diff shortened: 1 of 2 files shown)


ZergsLaw:
@googlebot I signed it!

mockgen/mockgen_test.go
```diff
@@ -267,3 +267,89 @@ func findMethod(t *testing.T, identifier, methodName string, lines []string) int
...
+				},
+			},
+			expected: []string{"firstArg", "arg1"},
+		},
+	} {
+		t.Run(testCase.name, func(t *testing.T) {
+			g := generator{}
+
+			result := g.getArgNames(testCase.method)
+			if !testEqSliceStr(t, result, testCase.expected) {
```
codyoss: I would recommend using `reflect.DeepEquals` here, but it seems like we have a name collision in that package. I will refactor this later

codyoss:
Thank you for your contribution! 🎆

mockgen/mockgen_test.go
ZergsLaw: Yes, that's what I planned initially, but since there was a conflict of names, I decided not to touch upon it at the moment.

PR #374, opened 2019-12-27
fix several linting warnings

codyoss:
**Description**

Found a bunch of warnings in my editor so I fixed them.

**Submitter Checklist**

review them:

- [x] Includes tests

**Reviewer Notes**

- [ ] Tests added.

**Release Notes**

N/A


mockgen/mockgen.go
```diff
@@ -305,14 +305,14 @@ func (g *generator) Generate(pkg *model.Package, pkgName string, outputPackagePa
 	g.p("")
 	g.p("import (")
 	g.in()
-	for path, pkg := range g.packageMap {
-		if path == outputPackagePath {
+	for pkgPath, pkg := range g.packageMap {
```
codyoss: We import `path` so we should not name variables with the same name as the import.

mockgen/mockgen.go
```diff
@@ -158,7 +158,7 @@ func parseMockNames(names string) map[string]string {
 }
 
 func usage() {
-	io.WriteString(os.Stderr, usageText)
+	_, _ = io.WriteString(os.Stderr, usageText)
```
codyoss: Being explicit.

mockgen/model/model.go
```diff
@@ -283,8 +284,8 @@ func (pt *PointerType) addImports(im map[string]bool) { pt.Type.addImports(im) }
 // PredeclaredType is a predeclared type such as "int".
 type PredeclaredType string
 
-func (pt PredeclaredType) String(pm map[string]string, pkgOverride string) string { return string(pt) }
-func (pt PredeclaredType) addImports(im map[string]bool)                          {}
+func (pt PredeclaredType) String(map[string]string, string) string { return string(pt) }
```
codyoss: If none of the parameters are used, don't name them.

gomock/controller_test.go
```diff
@@ -757,7 +757,7 @@ func TestVariadicNoMatch(t *testing.T) {
 	ctrl.RecordCall(s, "VariadicMethod", 0)
 	rep.assertFatal(func() {
 		ctrl.Call(s, "VariadicMethod", 1)
-	}, "Expected call at", "doesn't match the argument at index 0",
+	}, "expected call at", "doesn't match the argument at index 0",
```
codyoss: `error`s should not start with a capital letter

PR #378, opened 2020-01-03
fix tests passing when they should not

codyoss:


codyoss:
This PR now properly shows the generation failures that are happening. It appears with new modules stuff go generate does not behave the same if it is actually in the gopath, hence we were never seeing this failures. There must be some subtlety with generate that I am unaware of.

I would like to merge this with the failing CI, then make it top priority to fix these issues.

codyoss:
Pushing this through so it can unblock other issues.

PR #387, opened 2020-01-18
Use "." to refer to the current path's package in reflect mode

XSAM:
**Description**

I use `go:generate` in my source code to invoke `mockgen` to generate mocked interfaces.

e.g.

```go
//go:generate mockgen -destination=mocks/foo.pb.go github.com/xxx/xxx/pkg/xxx Foo
```

But it's too annoying that every time I have to write and check the package name (github.com/xxx/xxx/pkg/xxx).

And add the full package name in a source file is kind of duplicated information since *golang* could parse the file path and get the package name.

So I add a feature that can use "." to refer to the current path's package.

e.g.

```go
//go:generate mockgen -destination=mocks/foo.pb.go . Foo
```

Just a single `.`, no more `github.com/xxx/xxx/pkg/xxx`.

**Release Notes**

```
Use "." to refer to the current path's package in reflect mode
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,10 +48,15 @@ that uses reflection to understand interfaces. It is enabled
 by passing two non-flag arguments: an import path, and a
 comma-separated list of symbols.
 
+You can use "." to refer to the current path's package.
+
 Example:
 
 ```bash
 mockgen database/sql/driver Conn,Driver
+
+# Convenient for `go:generate`.
+mockgen . Conn,Driver
 ```
 
 The `mockgen` command is used to generate source code for a mock
--- a/mockgen/mockgen.go
+++ b/mockgen/mockgen.go
@@ -61,6 +61,7 @@ func main() {
 
 	var pkg *model.Package
 	var err error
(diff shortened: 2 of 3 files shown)


XSAM:
@googlebot I signed it!

codyoss:
Hey @XSAM, thanks for the PR. Is their a corresponding issue that goes with this PR? If not, could you please create one to discuss the feature.

XSAM:
I created an issue. #388

cvgw:
@XSAM  can we update this block https://github.com/golang/mock/blob/17ebec651714b485a91450f30c3b02d3cd068c2c/mockgen/mockgen.go#L262 so that the generated code has `// Source: {PKG_NAME}` instead of `// Source: .`

XSAM:
@cvgw sure, now it generates `// Source: {PKG_NAME}` while using `.` as package name.

cvgw:
@XSAM thank you for the contribution 🎉

PR #390, opened 2020-01-22
Reflect in Current Directory First

### LINKED ISSUE #354: mockgen always tries to pull latest version instead of required version(go.mod without vendor) in docker golang:1.13
docker image: golang:1.13
go version: go1.13.5
gomock version: all after v1.1.0
gomock work mode: reflect mode

However, in OSX, without docker, mockgen works well with go.mod. Very strange!

nguyenfilip:
Fixes #354

**Description**
Reflect in Current Directory First
  
Some users have intention to run mockgen in a module aware directory with
version pinned to those in `go.mod`.
    
Always try to run reflect program in current working directory first.

More context about the root cause of the bug is in https://github.com/golang/mock/issues/354#issuecomment-577147113

**Submitter Checklist**
- [ ] Includes tests

**Reviewer Notes**
- [ ] Tests added.

**Release Notes**

### DIFF
--- a/mockgen/reflect.go
+++ b/mockgen/reflect.go
@@ -129,6 +129,7 @@ func runInDir(program []byte, dir string) (*model.Package, error) {
 	if err := cmd.Run(); err != nil {
 		return nil, err
 	}
+
 	return run(filepath.Join(tmpDir, progBinary))
 }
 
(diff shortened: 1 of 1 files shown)


cvgw:
So previously this code tried to execute reflect in
1. The root directory of the pkg in question
2. The current directory that `mockgen` was execute from
3. The tmp directory

This change would make it
1. The current directory that `mockgen` was execute from
2. The root directory of the pkg in question
3. The tmp directory

Is that correct?

nguyenfilip:
@cvgw exactly. To be more precise the execution in `pkg` happenes only with module support. Without module support `mockgen` would try to first execute in GOPATH.

Issue #158, opened 2018-03-06
Mechanism of operation aborting valid codepath


ghost:
gomock is jumping out of a normal code execution, causing code that would normally be executed a normal program to be bypassed.  This means any clean-up after a failed mock call will have unexpected consequences such as:

- Files not being flushed or closed
- Signals not being sent over channels
- Mutexes being left in incorrect states

The last point is used as an example below.  The code being tested is a series of functions that use a global mutex for synchronization.

### Code to be tested:
```
package test

import "sync"
import "io"

var m sync.Mutex

func init() {
    m = sync.Mutex{}
}

func f(a io.Writer) {
    m.Lock()
    a.Write([]byte{})
    m.Unlock()
}

func g(a io.Writer) {
    m.Lock()
    a.Write([]byte{})
    m.Unlock()
}
```

### Test code
```
package test

import (
    "testing"

    "github.com/golang/mock/gomock"
)

func TestF(t *testing.T) {
    ctrl := gomock.NewController(t)
    defer ctrl.Finish()
    mw := NewMockWriter(ctrl)
    mw.EXPECT().Write([]byte{'a'})
    f(mw)
}

func TestG(t *testing.T) {
    ctrl := gomock.NewController(t)
    defer ctrl.Finish()
    mw := NewMockWriter(ctrl)
    mw.EXPECT().Write([]byte{'b'})
    g(mw)
}
```

### Command to generate mock:
```
mockgen -package test io Writer
```

### Output
```
➜  test go test .
--- FAIL: TestF (0.00s)
        controller.go:150: Unexpected call to *test.MockWriter.Write([[]]) at /home/ser/workspace/src/test/test_mock.go:32 because:
                Expected call at /home/ser/workspace/src/test/test_mock.go:39 doesn't match the argument at index 0.
                Got: []
                Want: is equal to [97]
        asm_amd64.s:573: missing call(s) to *test.MockWriter.Write(is equal to [97]) /home/ser/workspace/src/test/test_mock.go:39
        asm_amd64.s:573: aborting test due to missing call(s)
fatal error: all goroutines are asleep - deadlock!
(stack trace)
```

### Expected behavior
The mutex would in practice only be left in a locked state if the mocked functions panic'd, which is not the behavior being tested.  This results in incorrect execution of the tests, including in some cases tests hanging, tests being skipped, and an inability to properly execute tests.

[...]

PR #395, opened 2020-02-02
Fix #71 Do signature change error msg

### LINKED ISSUE #71: Hard to debug when a Do(fn) function no longer matches its mock signature
Situation: we had code like:
```go
type Foo interface {
  Bar(arg1, arg2 string)
  ...
}
```
and tests like:
```go
do := func(arg1, arg2 string) { ... }
// ... bunch of code ...
myBarMock.EXPECT().Bar(wantArg1, wantArg2).Do(do)
```

Then we added an argument to `Bar`:
```go
type Foo interface {
  Bar(arg1, arg2, arg3 string)
  ...
}
```
and updated the `EXPECT()`:
```go
do := func(arg1, arg2 string) { ... }
// ... bunch of code ...
myBarMock.EXPECT().Bar(wantArg1, wantArg2, wantArg3).Do(do)
```
... but forgot to update `do`. The result was "panic: reflect: Call with too many input arguments" with no strong indication that what was wrong was the `do` function, rather than some place where the third argument hadn't been added (e.g. missing `wantArg3`).
It would be nice if the message said something more obvious like "EXPECT.()Do() function doesn't match the mock signature".

cvgw:
Fixes #71 

**Description**

Update the error handling for Call.Do and Call.DoAndReturn in the case where
the argument passed does not match expectations.

* panic if the argument is not a function
* panic if the number of input arguments do not match those expected by Call
* panic if the types of the input arguments do not match those expected
by Call

Call.DoAndReturn has additional validations on the return signature

* panic if the number of return arguments do not match those expected by
Call
* panic if the types of return arguments do not match those expected by
Call

**Submitter Checklist**

review them:

- [x] Includes tests

**Reviewer Notes**

- [ ] Tests added.

**Release Notes**

```
- Update validations for Do + DoAndReturn; now panics when argument passed to Do is not a function or does not match the signature of the mocked function
```

### DIFF
--- a/gomock/call.go
+++ b/gomock/call.go
@@ -19,6 +19,8 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+
+	"github.com/golang/mock/gomock/internal/validate"
 )
 
 // Call represents an expected call to a mock.
@@ -105,9 +107,23 @@ func (c *Call) MaxTimes(n int) *Call {
 // DoAndReturn declares the action to run when the call is matched.
 // The return values from this function are returned by the mocked function.
 // It takes an interface{} argument to support n-arity functions.
+// If the method signature of f is not compatible with the mocked function a
+// panic will be triggered. Both the arguments and return values of f are
+// validated.
 func (c *Call) DoAndReturn(f interface{}) *Call {
-	// TODO: Check arity and types here, rather than dying badly elsewhere.
 	v := reflect.ValueOf(f)
+
+	switch v.Kind() {
+	case reflect.Func:
+		mt := c.methodType
+
+		ft := v.Type()
+		if err := validate.InputAndOutputSig(ft, mt); err != nil {
+			panic(fmt.Sprintf("DoAndReturn: %s", err))
+		}
+	default:
+		panic("DoAndReturn: argument must be a function")
+	}
 
 	c.addAction(func(args []interface{}) []interface{} {
 		vargs := make([]reflect.Value, len(args))
@@ -134,9 +150,23 @@ func (c *Call) DoAndReturn(f interface{}) *Call {
 // return values are ignored to retain backward compatibility. To use the
 // return values call DoAndReturn.
 // It takes an interface{} argument to support n-arity functions.
+// If the method signature of f is not compatible with the mocked function a
+// panic will be triggered. Only the arguments of f are validated; not the return
+// values.
 func (c *Call) Do(f interface{}) *Call {
-	// TODO: Check arity and types here, rather than dying badly elsewhere.
 	v := reflect.ValueOf(f)
+
+	switch v.Kind() {
+	case reflect.Func:
+		mt := c.methodType
+
+		ft := v.Type()
+		if err := validate.InputSig(ft, mt); err != nil {
+			panic(fmt.Sprintf("Do: %s", err))
+		}
+	default:
+		panic("Do: argument must be a function")
+	}
 
 	c.addAction(func(args []interface{}) []interface{} {
 		vargs := make([]reflect.Value, len(args))
--- a/gomock/call_test.go
+++ b/gomock/call_test.go
@@ -1,8 +1,30 @@
+// Copyright 2020 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package gomock
 
 import (
+	"fmt"
+	"reflect"
 	"testing"
 )
+
+type foo struct{}
+
+func (f foo) String() string {
+	return "meow"
+}
 
 type mockTestReporter struct {
 	errorCalls int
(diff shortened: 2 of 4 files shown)


cvgw:
Open question about this PR:
Should we verify the return args? Should we verify them for both `Do` and `DoAndReturn`?

gomock/call_test.go
```diff
@@ -1,6 +1,7 @@
 package gomock
 
 import (
+	"reflect"
```
codyoss: Could you add the copyright to this file. Eventually we need to go and get all the files caught up.

cvgw:
I did some brief testing and found that gomock currently allows arguments to `Do` which do not match the return signature of the mocked method. I don't think we should start validating that as it will be a break change so I'm going to revert that.

However, gomock raises an obscure error when arguments to `DoAndReturn` do not match the return signature of the mocked method. So I will keep the additional validation there.

gomock/internal/validate/validate.go
```diff
@@ -0,0 +1,201 @@
...
+			"expected function to have %d return vals not %d",
+			expectedFunc.NumOut(), actualFunc.NumOut())
+	}
+
+	for i := 0; i < expectedFunc.NumOut(); i++ {
+		expectedArg := expectedFunc.Out(i)
+		actualArg := actualFunc.Out(i)
+
+		if err := arg(actualArg, expectedArg); err != nil {
+			return errors.Wrapf(err, "return argument at %d", i)
```
codyoss: Do we need to wrap the errors? Could we append the old error as details? Or if we do want to wrap maybe we should consider `https://github.com/golang/xerrors/` instead as that is the api in go1.13. Thoughts?

cvgw: I don't think it's necessary to wrap them; I actually did it out of habit and raised the same question in chat. I recently read a great posts about how the errors you return are part of your public API, wrapping errors expands the surface of that API.

I'm happy to go with whatever you prefer.

codyoss: I would say for now let just not wrap since there is not a strong case for it. This has the benefit of pushing one less dependency on our users too.

cvgw: Done

gomock/call.go
```diff
@@ -106,9 +108,20 @@ func (c *Call) MaxTimes(n int) *Call {
...
+	switch v.Kind() {
+	case reflect.Func:
+		mt := c.methodType
+
+		ft := v.Type()
+		if err := validate.InputAndOutputSig(ft, mt); err != nil {
+			panic(fmt.Sprintf("DoAndReturn: %s", err))
+		}
+	default:
+		panic("DoAndReturn: argument must be a function")
```
codyoss: Since we are now explicitly panicing I think we should update the doc string.

gomock/internal/validate/validate.go
```diff
@@ -0,0 +1,199 @@
...
+			return false
+		}
+
+	}
+
+	return true
+}
+
+func interfaceArg(actualArg, expectedArg reflect.Type) error {
+	if !actualArg.ConvertibleTo(expectedArg) {
```
codyoss: Honest question: Does this cover the implements?

gomock/call.go
cvgw: Updated

gomock/internal/validate/validate.go
cvgw: Yes it does, if expectedArg is something like `Stringer` and `actualArg` is an instance of some type `Foo` which implements `Stringer` this line will evaluate to true.

cvgw: I added another test case to make that more clear

codyoss:
I was just reading over this issue. Made me question if we should be using panics at all. https://github.com/golang/mock/issues/158#issuecomment-378720709 

This PR is just making panics in this case more explicit. I wonder in the future if we should recover all of them and t.Fatal with the TestReporter  we should have access to in the controller.

cvgw:
> I was just reading over this issue. Made me question if we should be using panics at all. [#158 (comment)](https://github.com/golang/mock/issues/158#issuecomment-378720709)
> 
> This PR is just making panics in this case more explicit. I wonder in the future if we should recover all of them and t.Fatal with the TestReporter we should have access to in the controller.

I think that is totally valid but brings up back to the question of what to do for code executing outside of the main goroutine where we can't call `t.Fatal`. We should brainstorm about it some more.

horejsek:
BTW this change was released as 1.4.2, changing only minor version. It breaks the current code. I think it should be released at least as 1.5.0.

Linus-Boehm:
Can confirm that this is a breaking change. Gomock now interprets pointers wrong and wants a string:

Signature of func:

` Save(config *repo.Config, namespace domain.NS)`

Error:

`panic: DoAndReturn: input argument at 1: expected arg of kind string not interface`

codyoss:
Thank you for bring this up, that behavior was not intentional. Can you please provide a small example of code this breaks so we can add a better test.

horejsek:
```go
mock.EXPECT().Foo(gomock.Any()).DoAndReturn(func(...interface{}) error {return nil})
```

We didn't care about parameters so we left `...interface{}` there which was OK and now it's panicing. The change was easy thought, just use real type instead.

codyoss:
@Linus-Boehm was this a similar use case for you. I will be reverting for now and creating a new tag 1.4.3 until I have more time to evaluate how we want to move forward this this.

codyoss:
New release with the change reverted: [v1.4.3](https://github.com/golang/mock/releases/tag/v1.4.3)

Linus-Boehm:
`mock.EXPECT().Foo("some-expected-string").DoAndReturn(func(name interface{}) error {return nil})`

It panics becuase "some-expected-string" is not an interface, so yeah that should be the same reason.

cvgw:
> `mock.EXPECT().Foo("some-expected-string").DoAndReturn(func(name interface{}) error {return nil})`
> 
> It panics becuase "some-expected-string" is not an interface, so yeah that should be the same reason.

@Linus-Boehm for clarity, what is the signature of method `Foo` on the type being mocked?

PR #418, opened 2020-03-26
Using pacakges.NeedName

linzhp:
**Description**

`packages.LoadFiles` is [deprecated](https://pkg.go.dev/golang.org/x/tools/go/packages?tab=doc#pkg-constants). It is not future-proof to use it. In addition, I found all mockgen need is `Package.PkgPath`, so `packages.NeedName` is sufficient.

**Submitter Checklist**

review them:

- [x] Includes tests

**Reviewer Notes**

- [ ] Tests added.


codyoss:
We had talked about this internally at one time. @cvgw Brought this up, and I believe this should be safe.

codyoss:
Thank you for your contribution 🎆

PR #420, opened 2020-03-26
Avoid using packages.Load

### LINKED ISSUE #419: Avoid calling go list
@plakhotnii @codyoss 

`go list` is quite slow because it has to download external dependencies. This becomes more problematic under Bazel execution environment when GOCACHE is not available (https://github.com/bazelbuild/rules_go/issues/2366). I found all mockgen need from `go list` is the package path. Why not deduce the package path from source file or directory in mockgen, using logic like:

If GO111MODUE==off: 
   find the relative path from GOPATH
else:
  find the closest go.mod file and add the relative path from the go.mod file to the module name in that file
  if no go.mod exist and GO111MODUE != on:
    find the relative path from GOPATH

Related to #396

linzhp:
**Description**
Partially fixing #419 

It turn out there are two places calling `go list`:

* [createPackageMap](https://github.com/golang/mock/blob/0b87a54da2167cf3446363bb4b00c6de7677ceaa/mockgen/mockgen.go#L629)
* [parsePackageImport](https://github.com/golang/mock/blob/0b87a54da2167cf3446363bb4b00c6de7677ceaa/mockgen/parse.go#L552), which calls `packages.Load`, which calls `go list` by default.

This PR avoids calling `packages.Load` and saves one of the `go list` calls. Some performance data:

With 0b87a54da2167cf3446363bb4b00c6de7677ceaa:
```bash
$ go clean --cache --modcache && time go run ./mockgen -source /Users/zplin/gocode/src/github.com/bazelbuild/bazel-gazelle/language/lang.go > /dev/null
go: downloading golang.org/x/tools v0.0.0-20190425150028-36563e24a262
go: extracting golang.org/x/tools v0.0.0-20190425150028-36563e24a262
go: finding golang.org/x/tools v0.0.0-20190425150028-36563e24a262
go run ./mockgen -source  > /dev/null  6.98s user 6.10s system 121% cpu 10.767 total
```

After this PR:
```bash
$ go clean --cache --modcache && time go run ./mockgen -source /Users/zplin/gocode/src/github.com/bazelbuild/bazel-gazelle/language/lang.go > /dev/null
go: downloading golang.org/x/mod v0.2.0
go: extracting golang.org/x/mod v0.2.0
go: downloading golang.org/x/xerrors v0.0.0-20191011141410-1b5146add898
go: extracting golang.org/x/xerrors v0.0.0-20191011141410-1b5146add898
go: finding golang.org/x/mod v0.2.0
go: finding golang.org/x/xerrors v0.0.0-20191011141410-1b5146add898
go run ./mockgen -source  > /dev/null  3.77s user 3.81s system 118% cpu 6.383 total
```

**Submitter Checklist**

review them:

- [ ] Includes tests

**Reviewer Notes**

- [ ] Tests added.


linzhp:
@codyoss I will add tests if the overall approach looks good to you

codyoss:
Thanks for the PR, I probably won't have time to look at this today, but will try to get to it tomorrow. Thanks for being an active member in the community 😄

linzhp:
The travis-ci is failing even all tests are passing.

codyoss:
```bash
6.19s$ ./ci/check_go_generate.sh
prog.go:14:2: use of internal package github.com/golang/mock/mockgen/internal/tests/internal_pkg/subdir/internal/pkg not allowed
```

mockgen/parse.go
```diff
@@ -547,19 +547,44 @@ func packageNameOfDir(srcDir string) (string, error) {
...
-		return "", err
+	moduleMode := os.Getenv("GO111MODULE")
+	// trying to find the module
+	if moduleMode != "off" {
+		currentDir := srcDir
+		var modulePath string
+		for {
+			dat, err := ioutil.ReadFile(filepath.Join(currentDir, "go.mod"))
+			if os.IsNotExist(err) {
+				if currentDir == "." || currentDir == "/" {
```
codyoss: Does this work in Windows?

mockgen/parse.go
```diff
@@ -547,19 +547,44 @@ func packageNameOfDir(srcDir string) (string, error) {
...
+// pkgs, err := packages.Load(cfg, "file="+source)
+// However, it will call "go list" and slow down the performance
 func parsePackageImport(source, srcDir string) (string, error) {
-	cfg := &packages.Config{Mode: packages.LoadFiles, Tests: true, Dir: srcDir}
-	pkgs, err := packages.Load(cfg, "file="+source)
-	if err != nil {
-		return "", err
+	moduleMode := os.Getenv("GO111MODULE")
+	// trying to find the module
+	if moduleMode != "off" {
```
codyoss: Is `auto` a special case?

codyoss: Might be worth being explicit /w a switch

codyoss:
Overall, I think this seems good as long as we can test this throughly. Maybe we can/should also set different CI versions /w different GO111MODULE modes to help facilitate testing...

mockgen/parse.go
linzhp: the meaning of `auto` changes from Go release to release. To be safe, I only skip checking `go.mod` when the mode is "off". This may result in different behavior than "go list" in older Go releases.

For example, when "GO111MODULE" is "auto", if a directory is under GOPATH, but also has a go.mod file, "go list" will use GOPATH in 1.12, and use Go module in 1.13. This PR is similar to the behavior of 1.13.

codyoss:
Could you please re-pull master. I will get back to this review this week, sorry for the delay.

linzhp:
@codyoss any progress?

codyoss:
Sorry, was waiting for the CI to go green, but it appears to have gotten stuck. I will get to this today or tomorrow.

codyoss:
It looks like there are failures with go generate: https://travis-ci.org/github/golang/mock/jobs/672731180

linzhp:
@codyoss fixed CI errors

xtonyjiang:
It's too bad that in order to get this performance boost we have to not use `go/packages`, since I assume `go/packages` is supposed to allow us to not have to do all this branching logic based on whether we're using modules or not or what Go version we're using.

In the ideal world we would be able to take advantage of some `go/packages`-like thing that abstracts away the underlying import resolution system, without having the overhead of `go list`.

So I suppose we've accepted that we'll sacrifice possibly having to maintain this code path with future Go versions in order to get this speed boost? (Although maybe there won't be much more future maintenance because everything is stable now? Not sure what the Go roadmap looks like here.)

linzhp:
Yeah, Go's dependency manager is so tightly coupled with `go build`, making it very hard to reuse the dependency resolution logic if we use a different build system like Bazel.

Under Bazel's action environment, we don't have access to either GOCACHE or mod cache in the host machine. Every mockgen action runs from cold cache, similar to right after `go clean --cache --modcache`. Unfortunately, `go list` downloads and partially compiles the packages and dependencies. With cold cache, this is slow. The downloads and compilation are not necessary to find out the package path from a source directory. A lot of computation and I/O is wasted. 

Yes, there is a trade-off between performance and future maintenance. Since Bazel suffers from this issue the more than `go build`, I can potentially implement the logic in a package driver and set `GOPACKAGESDRIVER` in [bazel's gomock](https://github.com/jmhodges/bazel_gomock) rule so `packages.Load` will not call goListDriver.

I thought implementing it in golang/mock would benefit more people. That's why I sent PR here first.

linzhp:
@codyoss weekly ping...

linzhp:
@codyoss weekly ping...

codyoss:
@linzhp Thank you for your contribution! 🎆

PR #422, opened 2020-03-30
add default calling of ctrl.Finish() in go1.14+

### LINKED ISSUE #407: Use t.Cleanup (added in go1.14) to call `Finish` automatically
**Requested feature**

Currently, any test that uses `gomock` often has this snippet at the top:
```
defer ctrl.Finish()
```

The caller has to remember to `defer ctrl.Finish()`. This could be avoided if the gomock library supported go1.14's [`t.Cleanup`](https://golang.org/pkg/testing/#T.Cleanup) functionality.

**Why the feature is needed**
It reduces likelihood of the caller forgetting to call `Finish`, and reduces noise in tests.

**Proposed solution**
(Likely in a go1.14+-only file)
 * Make `Finish` OK to call multiple times.
 * Add a new constructor that takes `testing.TB` and calls `Finish` in a cleanup function

codyoss:
In go1.14 the testing package added a new cleanup method that is
called at the end of test runs. By taping into this gomock can
remove the need for users to call ctrl.Finish() if they pass a
*testing.T into gomock.NewController(...).

Fixes #407

### DIFF
--- a/gomock/controller.go
+++ b/gomock/controller.go
@@ -50,9 +50,6 @@
 //         mockObj.EXPECT().SomeMethod(2, "second"),
 //         mockObj.EXPECT().SomeMethod(3, "third"),
 //     )
-//
-// TODO:
-//	- Handle different argument/return types (e.g. ..., chan, map, interface).
 package gomock
 
 import (
@@ -75,6 +72,15 @@ type TestReporter interface {
 type TestHelper interface {
 	TestReporter
 	Helper()
+}
+
+// cleanuper is used to check if TestHelper also has the `Cleanup` method. A
+// common pattern is to pass in a `*testing.T` to
+// `NewController(t TestReporter)`. In Go 1.14+, `*testing.T` has a cleanup
+// method. This can be utilized to call `Finish()` so the caller of this library
+// does not have to.
+type cleanuper interface {
+	Cleanup(func())
 }
 
 // A Controller represents the top-level control of a mock ecosystem.  It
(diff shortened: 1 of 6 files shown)


gomock/controller_114_test.go
```diff
@@ -0,0 +1,132 @@
...
+	var ctrl *gomock.Controller
+	reporter.Cleanup(func() {
+		reporter.assertPass("Expected method call made.")
+	})
+	ctrl = gomock.NewController(reporter)
+	ctrl.RecordCall(subject, "FooMethod", "argument")
+	ctrl.Call(subject, "FooMethod", "argument")
+}
+
+func TestOrderedCallsInCorrect(t *testing.T) {
```
codyoss: With a couple of these tests I had to get a little fancy. Their implementation relied on Finish not being called to verify things. Cleanup acts in a LIFO matter. By registering a cleanup before the one in NewController, I can assert things after Finish() has been called.

gomock/controller.go
```diff
@@ -236,7 +263,15 @@ func (ctrl *Controller) Finish() {
...
 
 	if ctrl.finished {
-		ctrl.T.Fatalf("Controller.Finish was called more than once. It has to be called exactly once.")
+		if _, ok := isCleanuper(ctrl.T); !ok {
+			ctrl.T.Fatalf("Controller.Finish was called more than once. It has to be called exactly once.")
+		}
+		// provide a log message to guide users to remove `defer ctrl.Finish()` in Go 1.14+
+		tr := unwrapTestReporter(ctrl.T)
+		if l, ok := tr.(interface{ Log(args ...interface{}) }); ok {
+			l.Log("In Go 1.14+ you no longer need to `ctrl.Finish()` if a *testing.T is passed to `NewController(...)`")
```
codyoss: I can't decide if this is a good or bad idea. Do we let users know about this new behavior? I think this only shows up with `-v`. WDYT?

gomock/controller_113_test.go
```diff
@@ -0,0 +1,28 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build !go1.14
+
+package gomock_test
+
+import "testing"
+
+func TestDuplicateFinishCallFails(t *testing.T) {
```
codyoss: This was the previous behavior, lets keep things the same for versions less than 1.14

gomock/controller.go
cvgw: I think it's a good idea. I think you're correct that it would only show up if `-v` is passed

PR #444, opened 2020-06-09
feature/explicit-never

decanus:
**Description**

Although this can be done in other ways, it seems nice to have a more explicit method to ensure that a call never occurs. This adds the feature to do so.

**Submitter Checklist**

review them:

- [x] Includes tests

**Reviewer Notes**

- [ ] Tests added.

**Release Notes**

```
- Adds `Never` function for Calls
```

### DIFF
--- a/gomock/call.go
+++ b/gomock/call.go
@@ -82,6 +82,11 @@ func (c *Call) AnyTimes() *Call {
 	return c
 }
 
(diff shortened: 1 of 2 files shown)


codyoss:
Can you please open an issue for discussion and evaluation.

codyoss:
See notes on issue.

PR #452, opened 2020-06-26
fix docs for AssignableToTypeOf

ethervoid:
To make the example work we need to include the element to be casted

Fixes https://github.com/golang/mock/issues/453. _in case of a bug fix, this should point to a bug and any other related issue(s)_

**Description**

The example included in the documentation doesn't work. This PR fixes the documentation
to help other people that want to use it to learn how `AssignableToTypeOf` can be used.

**Submitter Checklist**

review them:

- [ ] Includes tests

**Reviewer Notes**

- [ ] Tests added.

**Release Notes**


ethervoid:
@googlebot I signed it!

PR #455, opened 2020-07-13
Fix https://github.com/golang/mock/issues/454

### LINKED ISSUE #454: Add the ability to add latency to mocks
I have some cases where time comes into play. A recent example:

```golang
// short-circuit early
if time.Now().After(deadline) {
    return errors.New("already too late")
}

// dep calls ... 

// short-circuit for product safety
if time.Now().After(deadline) {
    return errors.New("not enough time left")
}
```

I would like to be able to simulate latency during my mocked dep calls to facilitate testing this kind of use case.

### Workaround
Manually mock a dep and use `time.Sleep(x)` in the implementation. 

### Request
It would be nice to have something like `dep.EXPECT().Method().Latency(x).Return(nil)` where the recorder just sleeps for the duration I provide.

bradwilzon:
Fixes #454.

**Description**

Provide ability to add a latency (implemented as sleep time) to a Call.

**Submitter Checklist**

review them:

- [x] Includes tests

**Reviewer Notes**

- [ ] Tests added.

**Release Notes**

```
- Introduced a new API to Call.
```

### DIFF
--- a/gomock/call.go
+++ b/gomock/call.go
@@ -19,6 +19,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"time"
 )
 
 // Call represents an expected call to a mock.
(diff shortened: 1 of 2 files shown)


bradwilzon:
@googlebot I signed it!

codyoss:
See linked issue, closing.

PR #460, opened 2020-07-24
Fix empty error message when call is exhausted

### LINKED ISSUE #404: fix callserrors with empty msg
Check exhaustedNum and callsErrors.len() , if it's true, fill with error msg

There is a case where `matches` will return nil for an ordering error.
e.g.
```
gomock.InOrder(
  mockFoo.EXPECT().Foo().AnyTimes(),
  mockBar.EXPECT().Bar().AnyTimes(),
  mockBaz.EXPECT().Baz().AnyTimes(),
)

mockFoo.Foo()
mockBar.Bar()
mockFoo.Foo()
```
This will produce an error message with no additional information following `because: ` (as reported in the issue)

`err` will always be nil https://github.com/golang/mock/blob/1bf84fa8d4e5fedaedb9b103bb5b947e25bf5d8e/gomock/callset.go#L85

`len(exhausted)` will be 1 and `len(expected)` will be 0
https://github.com/golang/mock/blob/1bf84fa8d4e5fedaedb9b103bb5b947e25bf5d8e/gomock/callset.go#L90

_Originally posted by @cvgw in https://github.com/golang/mock/pull/292#issuecomment-569320658_

cvgw:
issue-404

Fixes #404 
**Description**

Fix empty error messages when the received call is exhausted

**Submitter Checklist**

review them:

- [x] Includes tests

**Reviewer Notes**

- [ ] Tests added.

**Release Notes**

```
Fix empty error message when expected method has been called too many times
```


gomock/callset.go
```diff
@@ -84,7 +84,11 @@ func (cs callSet) FindMatch(receiver interface{}, method string, args []interfac
 	for _, call := range exhausted {
 		if err := call.matches(args); err != nil {
 			_, _ = fmt.Fprintf(&callsErrors, "\n%v", err)
+			continue
 		}
+		_, _ = fmt.Fprintf(
+			&callsErrors, "all expected calls of the method %q have been exhausted for that received", method,
```
codyoss, line 90:
```suggestion
			&callsErrors, "all expected calls for method %q have been exhausted", method,
```

Issue #463, opened 2020-07-28
"go get" does not work anymore: issue with dependency rsc.io/quote/v3


olivierlemasle:
**Actual behavior** A clear and concise description of what the bug is.

`go get` this project does not work anymore if `rsc.io/quote/v3` is not cached and go proxy is not used.

**Expected behavior** A clear and concise description of what you expected to
happen.

`go get` this project should work.

**To Reproduce** Steps to reproduce the behavior

1.  Clear go mod cache
1.  Execute `go get` without go proxy:

```
$ GO111MODULE=on GOPROXY=direct go get github.com/golang/mock/mockgen@v1.4.3
go: downloading github.com/golang/mock v1.4.3
go: found github.com/golang/mock/mockgen in github.com/golang/mock v1.4.3
go get: github.com/golang/mock@v1.4.3 requires
	rsc.io/quote/v3@v3.1.0: unrecognized import path "rsc.io/quote/v3": reading https://rsc.io/quote/v3?go-get=1: 503 Service Unavailable

```
This is due to https://github.com/rsc/quote/issues/4 but it seems this dependency to `rsc.io/quote/v3` may be dropped, so this should be the solution if possible.

**Additional Information**

-   gomock mode (reflect or source): N/A
-   gomock version or git ref: `master`
-   golang version: `go1.14.3 linux/amd64`

**Triage Notes for the Maintainers**

codyoss:
I would recommend making using of the proxy if your environment allows. It accelerates the module downloads and can protect your builds against cases where hosts are unreliable.

I would use the default setting which is GOPROXY="https://proxy.golang.org,direct".

More details can be found here: https://golang.org/doc/go1.13#modules

codyoss:
I pushed `v1.4.4` that removes the need for this package.

helloworldcoding:
`import "rsc.io/quote/v4" `fixed my problem

PR #465, opened 2020-07-28
remove rsc.io/quote/v3 as dep

codyoss:


codyoss:
Fixes for GOPATH users on HEAD. Related #463

codyoss:
cc @cvgw

Issue #459, opened 2020-07-17
[Feature Request] Ability to check whether the expectations are met and reset mock expectations


hardikmodha:
### Requested feature: 
- **Ability to check whether all expectations are met** 
Currently, We only have `EXPECT()` which we can use in the user code and the rest (i.e. checking whether all expected calls were made) happens inside the controller. I would like to propose a utility which allows user code to check whether all expectations are met or not. A feature similar to the one which [go-sqlmock](https://godoc.org/github.com/DATA-DOG/go-sqlmock#SqlmockCommon) provides, namely `ExpectationsWereMet`
- **Resetting the mock behavior so that any previous behavior in not propagated to other tests** 
Currently, We need to redefine the `controller` to define the new mock behavior. I would like to propose a utility which allows user code to reset all previous expectations so that test separation is maintained and dirty state of other tests is not propagated to other tests. 
 
**Why the feature is needed** 
Right now, In my current project, While doing component-level testing, I need to inject the mocked dependencies of other services when my component gets initialized. So the controller is also defined there only. All these dependencies gets called in the code. It's time and resource consuming to initialize the application on each and every test. So if the above features are added then it would really help. 

<details>
<summary>The following is the high-level code structure on how I'm planning on consuming these features
</summary>

For the sake of brevity I've included only high-level part.
```go
file: ct_main_test.go // This defines the TestMain.go and initializes the application

var remoteService

func TestMain(m *testing.M) {
    // initialize application
    ctrl := controller.NewController(helper)
    remoteService = mock.MockedRemote(ctrl)
    service = initializeMyService(remoteService)    
    .
    .
    os.Exit(m.Run())
}

file: ct_some_file_test.go

func Test_xxx(t *testing.T) {
      // Before proceeding, Reset any previous expectations
      remoteservice.ResetExpectations()

      // Given
      remoteService.Expect().ServiceMethod().Times(2)

     // When an API is called
     myService.SayHello()

     // Then
     remoteService.ExpectationsWereMet()    // Assert that all expectations were met. If expectations were not met, we throw an error
}
```
</details>

This would really help in implementing best practices in my current project and will help in avoid passing the dirty state between tests mistakenly.

**PS**: Excuse me if it's possible somehow using the current features provided by `gomock`. In that case, Let me know how can I achieve the same. Thanks!

cvgw:
Making sure I understand this request;

Your tests require a lot of set up, particularly in the mocking of expectation?
Your tests re-use a lot of the same setup?
[...]

PR #467, opened 2020-08-12
add reset method

DrWrong:
Add a `Reset` method for cases where I have to reuse `Controller` 

**Description**

when use `gomock` for a integration testing scenario, where the testing programing depends on lots of grpc interface, I can not just renew a Controller for each case , so a `reset` method is needed

**Submitter Checklist**

review them:

- [ ] Includes tests

**Reviewer Notes**

- [ ] Tests added.

**Release Notes**

```
- Improved API for custom matchers
```

### DIFF
--- a/gomock/controller.go
+++ b/gomock/controller.go
@@ -256,6 +256,13 @@ func (ctrl *Controller) Call(receiver interface{}, method string, args ...interf
 	return rets
 }
 
+// Reset rest expectedCalls
+func (ctrl *Controller) Reset() {
+	ctrl.mu.Lock()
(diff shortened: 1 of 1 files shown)


DrWrong:
@googlebot I signed it!

cvgw:
This seems similar to https://github.com/golang/mock/issues/459 which is currently under discussion

codyoss:
Closing this PR for now. I would like to discuss more on the issue before we implement this, also we would need to add some tests should this feature be accepted. Thanks for understanding.

PR #470, opened 2020-08-14
Add example for Call.Do and Call.DoAndReturn

### LINKED ISSUE #469: Improve documentation for `Do` and `DoAndReturn`
The docs for `Do` and `DoAndReturn` are not as clear as they could be and this has been a source of confusion/github issues.

We should improve the docs for both methods including adding clear examples to the GoDoc

cvgw:
Fixes #469 

**Description**

Add examples for `Call.Do` and `Call.DoAndReturn`

**Submitter Checklist**

review them:

- [ ] Includes tests

**Reviewer Notes**

- [ ] Tests added.


gomock/doc_test.go
```diff
@@ -0,0 +1,66 @@
...
+			return "I'm sleepy"
+		},
+	)
+}
+
+func ExampleCall_DoAndReturn_captureArguments() {
+	t := &testing.T{} // provided by test
+	ctrl := gomock.NewController(t)
+	mockIndex := mock_sample.NewMockIndex(ctrl)
+	var s string
```
codyoss, line 29: Not used. Lets assign arg to this value and print it. Then we can make this a tested example. See [examples blog](https://blog.golang.org/examples)

codyoss: Same for Do.

gomock/doc_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+	mockIndex.EXPECT().Get(gomock.AssignableToTypeOf(s)).DoAndReturn(
+		// When capturing arguments the anonymous function should have the same signature as the mocked method.
+		func(arg string) interface{} {
+			time.Sleep(1 * time.Second)
```
codyoss, line 34: nit: remove for capture example. Also, maybe make Millisecond if these will be run with tests to keep things snappy.

gomock/doc_test.go
```diff
@@ -0,0 +1,66 @@
...
+			fmt.Println(arg)
+			return "I'm sleepy"
+		},
+	)
+}
+
+func ExampleCall_Do() {
```
codyoss, line 41: call out this is a latency example in the name?

PR #474, opened 2020-08-18
Introduce ctrl.Continue() to allow reuse of controller

### LINKED ISSUE #459: [Feature Request] Ability to check whether the expectations are met and reset mock expectations

hardikmodha:
Closes #459.

**Description**

Introduce a new construct `Continue` on the controller. It will be almost like `ctrl.Finish` with the following differences.

1. It can be called multiple times from the user code.
2. It will implicitly reset the calls at the end.

Above will ensure that when the user code calls `ctrl.Continue()` all the expectations are checked for and calls are also reset. Now, user code can reuse the same controller to define new expectations. Older expectations are no longer in the picture. This way we can be sure that unmet expectations of any tests don't have side-effects on other tests

**Submitter Checklist**

- [x] Includes tests

**Reviewer Notes**

- [ ] Tests added.

### DIFF
--- a/gomock/controller.go
+++ b/gomock/controller.go
@@ -256,6 +256,13 @@ func (ctrl *Controller) Call(receiver interface{}, method string, args ...interf
 	return rets
 }
 
+// ResetCalls - Reset expectedCalls
+func (ctrl *Controller) ResetCalls() {
+	ctrl.mu.Lock()
+	defer ctrl.mu.Unlock()
+	ctrl.expectedCalls = newCallSet()
+}
+
 // Finish checks to see if all the methods that were expected to be called
 // were called. It should be invoked for each Controller. It is not idempotent
 // and therefore can only be invoked once.
--- a/gomock/controller.go
+++ b/gomock/controller.go
@@ -256,9 +256,9 @@ func (ctrl *Controller) Call(receiver interface{}, method string, args ...interf
 	return rets
 }
 
-// ResetExpectedCalls Reset expectedCalls so that we can reuse the same controller to define
+// Reset expectedCalls so that we can reuse the same controller to define
 // new set of expected calls. See https://github.com/golang/mock/issues/459
-func (ctrl *Controller) ResetExpectedCalls() {
+func (ctrl *Controller) reset() {
 	ctrl.mu.Lock()
 	defer ctrl.mu.Unlock()
 	ctrl.expectedCalls = newCallSet()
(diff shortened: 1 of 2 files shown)


hardikmodha:
@googlebot I signed it!

codyoss:
Closing this for reasons discussed in the linked issue.

Issue #428, opened 2020-04-23
panic in test is swallowed if there are failed expectations and t.Cleanup(ctrl.Finish)


powerman:
**Actual behavior** A clear and concise description of what the bug is.

If test function setup some `EXPECT()` which wasn't called yet and then it `panic()`, the panic message won't be shown in test output, all I get is `missing call(s) to … aborting test due to missing call(s)` message.

**Expected behavior** A clear and concise description of what you expected to happen.

Panic has to be shown in test output. Mentioned above message about missing calls may or may not be shown, it's much less important than panic and often implicit because of panic.

**To Reproduce** Steps to reproduce the behavior

I can provide trivial example if needed, but I believe it's obvious.

**Additional Information**

-   gomock mode (reflect or source): source
-   gomock version or git ref: v1.4.3
-   golang version: go version go1.14.2 linux/amd64

**Triage Notes for the Maintainers**

codyoss:
Thanks for the feedback, I agree I think we could do more for the cases where the mocks are not quite setup right. I think there are other related issues here, will link when I re-come across them.

lassekv:
I just ran into this issue today as well. This only happens from Go 1.14 and forward when an explicit defer ctrl.Finish() is no longer expected because it is registered through the testing.T interface. If one explicitly does a defer ctrl.Finish() then the panics are not swallowed.

powerman:
This is still quite annoying - test just hangs without any output. @codyoss If you can provide some quick hints about how to implement possible fix then I can try to prepare PR.

In the meanwhile I'd like to share two workarounds:
1. Run `go test -timeout=5s` to work around hang and get some details about which test is broken/panics.
2. Handle panic manually inside tests to actually workaround this issue (but it's inconvenient because this code must be added inside every Test-function and also inside every t.Run-function):

```go
defer func() {
	if err := recover(); err != nil {
		t.Error(err)
	}
}()
```

codyoss:
@powerman I did open up a PR for this yesterday. You can see the link above your comment 😸 I think that should solve the issue. Check out the PR and let me know if it does not though.

powerman:
@codyoss Your PR didn't fix the issue. Here is minimal example to reproduce hang:

```go
//go:generate mockgen -package=$GOPACKAGE -source=$GOFILE -destination=mock.$GOFILE T

package gomock428

import (
	"testing"

	"github.com/golang/mock/gomock"
)

type T interface {
	F(string) error
}

[...]

PR #478, opened 2020-09-01
surface panic when calling Finish is implicit

codyoss:
When Finsh is called from Cleanup, it is not done so in a deffered
context. Because of this, we can't recover from panics. The
testing package will re-surface any panics but only after the
Cleanup functions is executed. Because we were calling t.Fatal we
were hiding panics from users of the library when they were using
Finish implicitly. Fatal calls runtime.Goexit() under the hood so
the panic never gets the chance to be handled by the testing
package.

Also, removed logging as after thinking about it more it seems a
little aggressive to push the feature onto users.

Note, when there is a panic and Finish is called from the Cleanup
function that a few extra lines will be logged from the calls to
ErrorF. This seems like a small price to pay to get the context
of the panic

Fixes: #428


gomock/controller.go
```diff
@@ -260,6 +260,13 @@ func (ctrl *Controller) Call(receiver interface{}, method string, args ...interf
 // were called. It should be invoked for each Controller. It is not idempotent
 // and therefore can only be invoked once.
 func (ctrl *Controller) Finish() {
+	// If we're currently panicking, probably because this is a deferred call.
+	// This must be recovered in the deferred function.
+	err := recover()
+	ctrl.finish(false, err)
+}
+
+func (ctrl *Controller) finish(cleanup bool, panicErr interface{}) {
```
cvgw, line 269: can panicErr just be type `error`?

cvgw (approved):

gomock/controller.go
codyoss: It can't, it could be anything. It represents whatever what the input to `panic`.: https://github.com/golang/go/wiki/PanicAndRecover#panic

cvgw: Ah ok, makes sense

Issue #429, opened 2020-04-24
EXPECT fails when using type derived from map


jtyers:
If I derive a type from a core type and use that in my expectations, tests start failing where they didn't before, even where the inputs have not changed. I've only tested this with a `map`.

Below is a complete working test that shows a given input working when I pass a `map` to the EXPECT (and mock), and then when I pass an `M{}`. The inputs stay the same, but the second test fails.

```
package main

//go:generate mockgen -source=$GOFILE -destination=test_mock.go -package main

import (
	"github.com/golang/mock/gomock"
	"testing"
)

type M map[string]interface{}

type TestInterface interface {
	DoSomething(input map[string]interface{})
}

func TestDoSomething(t *testing.T) {

	t.Run("using maps", func(t *testing.T) {
		ctrl := gomock.NewController(t)
		defer ctrl.Finish()

		mock := NewMockTestInterface(ctrl)

		mock.EXPECT().DoSomething(map[string]interface{}{"foo": "bar"})

		mock.DoSomething(map[string]interface{}{"foo": "bar"})

	})

	t.Run("using M", func(t *testing.T) {
		ctrl := gomock.NewController(t)
		defer ctrl.Finish()

		mock := NewMockTestInterface(ctrl)

		mock.EXPECT().DoSomething(M{"foo": "bar"})

		mock.DoSomething(M{"foo": "bar"})

	})
}
```

**Additional Information**
Golang 1.14.1

codyoss:
@jtyers  Can you please share the versions were this both did and did not work(mockgen)?

jtyers:
@codyoss sorry, here it is:

```
go: found github.com/golang/mock/mockgen in github.com/golang/mock v1.4.3
go: downloading golang.org/x/tools v0.0.0-20190425150028-36563e24a262
```

jtyers:
To be clear I'm not aware of this ever working in mockgen. My reference to "tests failing where they didn't before" is that in the old version of my code, I used `map[string]interface{}` and in the new version, `M`.

jtyers:
Also of interest may be that if I change the signature to `DoSomething(input M)` the test stlil fails.

codyoss:
Thank you for all the information that helps! Yes is seems like our type checking might be too strict. My guess with out looking at the code is we need to add some reflect checks around if the type is assignable. I would be glad to accept a PR for such a feature 😸

PR #481, opened 2020-09-23
MOCK-429: add support for assignable types to Eq matcher

GrigoriyMikhalkin:
Solution to https://github.com/golang/mock/issues/429. Adds support for assignable types to `Eq` matcher.

### DIFF
--- a/gomock/matchers.go
+++ b/gomock/matchers.go
@@ -102,7 +102,21 @@ type eqMatcher struct {
 }
 
 func (e eqMatcher) Matches(x interface{}) bool {
-	return reflect.DeepEqual(e.x, x)
+	// In case, some value is nil
+	if e.x == nil || x == nil {
+		return reflect.DeepEqual(e.x, x)
+	}
+
+	// Check if types assignable and convert them to common type
+	x1Val := reflect.ValueOf(e.x)
+	x2Val := reflect.ValueOf(x)
+
+	if x1Val.Type().AssignableTo(x2Val.Type()) {
+		x1ValConverted := x1Val.Convert(x2Val.Type())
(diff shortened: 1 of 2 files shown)


cvgw:
Hi @GrigoriyMikhalkin, is this PR ready for review?

GrigoriyMikhalkin:
@cvgw Hi! Yep.

gomock/matchers.go
```diff
@@ -102,7 +102,21 @@ type eqMatcher struct {
...
+	// In case, some value is nil
+	if e.x == nil || x == nil {
+		return reflect.DeepEqual(e.x, x)
+	}
+
+	// Check if types assignable and convert them to common type
+	x1Val := reflect.ValueOf(e.x)
+	x2Val := reflect.ValueOf(x)
+
+	if x1Val.Type().AssignableTo(x2Val.Type()) {
```
cvgw, line 114: I think for aliased types (`type A []string`) it doesn't matter, but if we are talking about interfaces I think there may be an issue here.

```
type Foo interface {
  Meow() string
}

type Baz struct {}

func (b Baz) Meow() string {
  return "meow"
}

type Boomer interface {
  Boom(Foo)
}

func TestBoom(t *testing.T) {
  // omitted for brevity
  var bazzer Baz = newBazzer()
  mockBoomer.EXPECT().Boom(gomock.Eq(bazzer)).Times(1)
  var fooer Foo = Foo(bazzer)
  mockBoomer.Boom(fooer) // Foo not assignable to Baz
}
```

cvgw: Ah ok, I think I understand. In the case of an interface the object will be resolved to its underlying type which must be assignable to the expected type. Thanks for the demo!

cvgw (approved):

Issue #351, opened 2019-11-27
Custom Matcher


Menouer:
Hi,

I want to validate only subfields of an object, sadly i did not find a way to do this using gomock. Is there  a way to create a custom matcher ? for example passing a fonction `f(interface...)  bool` that will contain the logic ?

If not is there any way i can do this using gomock ? :)

Thanks !

Menouer:
Given the added tags i assume i was not clear in my first post. Let's try again :D.

Let's say i have a struct, this struct has multiple fields and is going to be passed to a method from a mocked service. But sadly within this struck some of the fields are generated "randomly" by another service that is not mocked and should not be. 

Now what i would like to do if possible is to validate only a subset of this struct and ignore the "randomly" generated fields.

Is this possible ? 

Thank you ! :)

rubens21:
It may be useful to you,

```go
import "github.com/golang/mock/gomock"

func NewMatcher(customMatcher func(arg interface{}) bool) gomock.Matcher {
	return matcherCustomizer{customMatcher}
}

type matcherCustomizer struct {
	matcherFunction func(arg interface{}) bool
}

func (o matcherCustomizer) Matches(x interface{}) bool {
	return o.matcherFunction(x)
}

func (o matcherCustomizer) String() string {
	return "[call back function matcher has returned false]"
}

```

If you need an accurate error message, you may improve it to store an error message after the function be called.

Menouer:
That is exactly what i wanted, Thanks you ! :)

codyoss:
Thanks @rubens21 for helping out!

jtyers:
Stumbled across this issue while looking for a way to do custom matchers. @codyoss could this be added to the README? Would make it much easier to find.

PR #496, opened 2020-10-28
Adds a customizable matcher

rubens21:
Helps cases like #351. 

**Description**

This PR creates a customizable matcher that uses a custom function to evaluate the value. 

**Release Notes**

Improved API for custom matchers 

`I know that release note was an example, but that was exactly what this PR did!`

### DIFF
--- a/gomock/matchers.go
+++ b/gomock/matchers.go
@@ -208,6 +208,27 @@ func (m lenMatcher) String() string {
 	return fmt.Sprintf("has length %d", m.i)
 }
 
+type customMatcher struct {
+	matcherFunction func(arg interface{}) error
+	err             error
+}
+
+func (o *customMatcher) Matches(x interface{}) bool {
+	// in case of this matcher has been used in a failed test before
+	o.err = nil
+	if err := o.matcherFunction(x); err != nil {
+		o.err = err
+	}
+	return o.err == nil
+}
+
+func (o *customMatcher) String() string {
+	if o.err == nil {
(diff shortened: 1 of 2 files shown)


rubens21:
@googlebot I signed it!

codyoss:
@rubens21 Thanks for the PR, but this is not actually necessary. [Matcher](https://pkg.go.dev/github.com/golang/mock/gomock#Matcher) is already an interface that anyone can implement. You can define your own matchers in our codebase today and pass them in.

If you think this is not known today I think maybe we need to add some more/better docs around this. Can you open up an issue for that?

rubens21:
> @rubens21 Thanks for the PR, but this is not actually necessary. [Matcher](https://pkg.go.dev/github.com/golang/mock/gomock#Matcher) is already an interface that anyone can implement. You can define your own matchers in our codebase today and pass them in.
> 
> If you think this is not known today I think maybe we need to add some more/better docs around this. Can you open up an issue for that?

Hi @codyoss , yeah I know anyone can implement that interface, and sometimes it is not a big deal.

However, many times implementing an interface (struct and two methods) to test a simple field is a painful work for a simple task. Having a shortcut to simplify that to a function could be handy.

Please consider to re-think about the advantage vs disadvantage of merging that. I have been used gomock for a couple of years already, and I used that matcher in almost all my projects.

codyoss:
I have thought about this and think this is better solved by documentation.

PR #500, opened 2020-11-04
sort methods on the mock implementation alphabetically

### LINKED ISSUE #499: inconsistent order of methods on the mock interface between source and reflect mode
**Actual behavior**
Consider the following interface:

```go
type WriteCloser interface {
	Write()
	Close()
}
```
In source mode `Write()` will come before `Close()`. In reflect mode, `Close()` will come before `Write()`.

While the Go compiler doesn't care about the order, it would be nice if the order was consistent between the two modes (imagine switching from source to reflect mode in a project. It would be nice if you wouldn't have to commit a large batch of changes to the mock files in your VCS).

**Expected behavior** 

The order should be consistent. Alphabetically ordered, for example.

see example

**Additional Information**

-   gomock mode (reflect or source): both
-   gomock version or git ref: 1.4.4
-   golang version: 1.15.3

marten-seemann:
Fixes #499.

**Description**

Sorting the methods by name will make sure that their order is consistent in source and reflect mode.

**Submitter Checklist**

review them:

- [x] Includes tests

Updated the tests, but not sure how to test this without writing a parser for the generated mock files.

**Reviewer Notes**

- [ ] Tests added.

### DIFF
--- a/mockgen/internal/tests/aux_imports_embedded_interface/bugreport_mock.go
+++ b/mockgen/internal/tests/aux_imports_embedded_interface/bugreport_mock.go
@@ -34,20 +34,6 @@ func (m *MockSource) EXPECT() *MockSourceMockRecorder {
 	return m.recorder
 }
 
-// Method mocks base method.
-func (m *MockSource) Method() faux.Return {
-	m.ctrl.T.Helper()
-	ret := m.ctrl.Call(m, "Method")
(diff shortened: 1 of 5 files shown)


marten-seemann:
@codyoss Any thoughts on this PR?

codyoss (approved):

Issue #505, opened 2020-12-21
Missing import when -package with same package as the destination 


keisar:
**Actual behavior**
Assuming we have the following project structure:
* /source/source.go -> package name is "pkg1" (import for this package is "source/pkg1")
* /output/empty.go -> package name is "pkg1"  (import for this package is "output/pkg1")
When running `mockgen -package pkg1 -destination ../output/source_mock.go -source=source.go` the generated mock is created without imports to "source/pkg1" and the mock has syntax errors.

**Expected behavior** 
The generated mock to include an import to "source/pkg1" and not have syntax errors.

**To Reproduce** Steps to reproduce the behavior
You can use [this test](https://github.com/keisar/mock/blob/master/mockgen/internal/tests/import_source2/definition/source.go) to reproduce the issue

**Additional Information**
I believe this used to work fine in version 1.3.x, I couldn't pinpoint exactly when this started happening.
I am willing to contribute a fix but will need a little help/guidance as where to start looking, I started looking at how the packages are mapped but couldn't find where this might be happening.

-   gomock mode (reflect or source): source
-   gomock version or git ref: 1.4.x
-   golang version: 1.13 & 1.15

**Triage Notes for the Maintainers**

keisar:
From debugging the test I mentioned above I believe the issue lies [here ](https://github.com/golang/mock/blob/master/mockgen/mockgen.go#L333), it seems `pkg.PkgPath` has the source package path and not the output package path, I'll take a closer look, if anyone can confirm this I will try to submit a fix.

keisar:
I believe the issue is a combination of two problems:
1. Building `outputPackagePath` ([here](https://github.com/golang/mock/blob/master/mockgen/mockgen.go#L136)) doesn't work for go modules as it relies on SrcDir().
2. [here](https://github.com/golang/mock/blob/master/mockgen/mockgen.go#L333) it should use `outputPackagePath` instead of `pkg.PkgPath` and if `outputPackagePath` is empty we should set it to `pkg.PkgPath` at the start of `Generate`

From my testing fixing these two issues keeps existing functionality and fixes the issue above, I couldn't find a good way of generating `outputPackagePath` yet so for my testing I set it manually to the correct path.

If you agree with my findings I can try and come up with a solution for generating `outputPackagePath` and send a PR.

codyoss:
[...]

PR #507, opened 2021-01-15
fix issues related to source package imports

codyoss:
Fixes: #505
Fixes: #515


codyoss:
@cvgw Could you take a look at this when you have time 😄

PR #512, opened 2021-01-22
fix Test_createPackageMap for 1.15

codyoss:
In Go 1.15 go list will return an error if any of the packages
passed to it are not valid. This differs from pervious behavior
where all of the successful results would be returned. In practice
this should not be an issue as code should contain resolvable
import paths, but this made the test fail.


codyoss:
Merging to get things back to green to unblock other PRs.

PR #529, opened 2021-02-24
allow for overriding source package

choo-stripe:
Greetings!

We call `mockgen` from bazel, which lives in a hermetic environment and doesn't have access to go modules or GOPATH. However, we shouldn't really need to magically find the name of the go package.

This PR adds a flag to override the source package - so that if the mock file needs to import from the original file, it'll use the provided package.

I looked at writing tests. I didn't put this in `parsePackageImport` because that function is called in multiple places, including to determine the destination path. We don't want to override the destination path. I thought about creating a new function, `getPackageName` or something, but it didn't seem worthwhile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Alternatively, there is an online reference for the package hosted on GoPkgDoc
 
 Source mode generates mock interfaces from a source file.
 It is enabled by using the -source flag. Other flags that
-may be useful in this mode are -imports and -aux_files.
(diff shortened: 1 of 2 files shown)


codyoss:
@choo-stripe Would you please open a feature request for this PR. We like to have discussion before implementing any features.

choo-stripe:
Done: https://github.com/golang/mock/issues/531

codyoss:
Closing in favor of #547

PR #544, opened 2021-03-22
test(sample/user_test.go): minor correction at t.Errorf

adsian:
test(sample/user_test.go): minor correction at t.Errorf: format


codyoss (approved):
Good catch, thanks!

Issue #545, opened 2021-03-22
Built-in matcher that compares slices of arbitrary elements order


vvkh:
# Requested feature
It would nice to have built-in matcher that compares slices and returns `true` if slices have same elements in arbitrary order.

# Why the feature is needed
Sometimes code we are testing produces slices with non-determined elements order. For example, when converting map to slice we may end up with many possible orders of elements.
```
s := make([]interface{}, 0, len(m))
for k, v := range m {
    s = append(s, v)
}
```
Some testing frameworks like testify include [special assert function](https://github.com/stretchr/testify/blob/6990a05d54c2287be8787a411d77815643347339/assert/assertions.go#L895) checking if two slices have same elements.

If we have a mock depending on arguments with non-determined elements order such as `s` then setting up expected calls for such mock can be quite inconvenient. We either have to use `Any()` matcher and skip validating this argument at all or implement custom matcher.  

# Proposed solution
Include built-in matcher like `InAnyOrder(s []interface{})` or `WithSameElements(s []interface{}` implementing checks similar to [assert.ElementsMatch](https://github.com/stretchr/testify/blob/6990a05d54c2287be8787a411d77815643347339/assert/assertions.go#L895)

P.S.
I will be happy to implement the matcher if we agree on this feature being useful.

codyoss:
Hey, thanks for the feature request. I will let this issue sit for a little bit to see what others in the community think of the idea. I could see how a matcher like this could be useful.

vvkh:
Hey @codyoss.  I've provided [draft implementation](https://github.com/golang/mock/pull/546) for requested issue. Could you please take a look?

vvkh:
Implemented in https://github.com/golang/mock/pull/546

PR #546, opened 2021-03-25
feat: add InAnyOrder matcher

vvkh:
Draft implementation for #545


gomock/matchers.go
```diff
@@ -207,6 +207,66 @@ func (m lenMatcher) String() string {
...
+	given, ok := m.prepareValue(x)
+	if !ok {
+		return false
+	}
+	wanted, ok := m.prepareValue(m.x)
+	if !ok {
+		return false
+	}
+
+	usedFromGiven := make([]bool, given.Len())
```
codyoss, line 224: Maybe we should fail fast here if lens don't match, they should be exact right?

gomock/matchers.go
```diff
@@ -207,6 +207,66 @@ func (m lenMatcher) String() string {
...
+		}
+	}
+
+	return extraInGiven == 0 && missingFromWanted == 0
+}
+
+func (m inAnyOrderMatcher) prepareValue(x interface{}) (reflect.Value, bool) {
+	xValue := reflect.ValueOf(x)
+	switch xValue.Kind() {
+	case reflect.Slice, reflect.Array, reflect.String:
```
codyoss, line 259: I am not sure it makes sense here to also allow strings. I think we should just stick to arrays and slices for now.

gomock/matchers_test.go
```diff
@@ -142,3 +142,176 @@ func TestAssignableToTypeOfMatcher(t *testing.T) {
...
+
+func TestInAnyOrder(t *testing.T) {
+	tests := []struct {
+		name      string
+		wanted    interface{}
+		given     interface{}
+		wantMatch bool
+	}{
+		{
+			name:      "match_for_equal_slices",
```
codyoss, line 154: nit: can use spaces instead of underscores for test names.

codyoss (commented):
Just a couple of small things, thanks for the contribution!

gomock/matchers.go
vvkh: Good point, added early return for this case.

gomock/matchers.go
vvkh: Ok, we can add it later if someone actually needs this. 👌

gomock/matchers_test.go
vvkh: Sure, it's just nice to be able to copy name of a test with underscores directly from gotest log as is and search for it in codebase. :)

codyoss: Looks like this was not resolved.

codyoss (changes requested):
Just one minor thing, other than that looks good to me!

codyoss (approved):

PR #547, opened 2021-03-27
Best effort guesses for output package path

linzhp:
I know we need output package path to avoid self-importing. However, this only happens when the mock is output into an already existing package, in which case the caller is responsible for passing `-self_package`. Without `-self_package`, guessing output package path should be on best effort basis. There is no requirement for `destination` to be in a Go module or `GOPATH` (in fact, it's hard, if possible at all, to do so in Bazel). So `parsePackageImport(dstPath)` often fails. Such failure should not fail the main program. We should instead be optimistic and leave output package package empty, assuming no self importing would happen.


codyoss:
This seems reasonable to me, thanks for the contribution!

codyoss (approved):

PR #556, opened 2021-05-13
Add statement in Finish function about there is no need to call Finish() 

sanposhiho:
In addition to the description in NewController function's docs, it would be nice if there was a statement in Finish function's docs that there is no need to call Finish() under specific circumstances.


codyoss (approved):
LGTM, thank you!

PR #558, opened 2021-05-14
feat validate Do & DoReturn args

codyoss:
Previous behavior was a panic would happen if the number of args
did not match. Now the test should fail more gracefully and allow
for better debugging.

Fixes: #71

### DIFF
--- a/gomock/call.go
+++ b/gomock/call.go
@@ -113,8 +113,14 @@ func (c *Call) DoAndReturn(f interface{}) *Call {
 	v := reflect.ValueOf(f)
 
 	c.addAction(func(args []interface{}) []interface{} {
+		c.t.Helper()
 		vArgs := make([]reflect.Value, len(args))
 		ft := v.Type()
+		if c.methodType.NumIn() != ft.NumIn() {
+			c.t.Fatalf("wrong number of arguments in DoAndReturn func for %T.%v: got %d, want %d [%s]",
+				c.receiver, c.method, ft.NumIn(), c.methodType.NumIn(), c.origin)
+			return nil
+		}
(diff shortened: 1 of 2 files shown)


bconway:
Apologies, my reflection-fu is a little weak. Does this correctly handle variadic functions? I have a bunch of `DoAndReturn` with throw-away input in the form of `.DoAndReturn(func(_ ...interface{}) error` that worked without issue in 1.5.0 and now throw:

```
wrong number of arguments in DoAndReturn func for *notifier.Mockeventer.Create: got 1, want 2
```

in 1.6.0.

Or perhaps this is a Hyrum’s Law thing?

codyoss:
@bconway Would you please open an issue with a simple example for discussion. This may be too strict in the case of varargs... need to take a closer look.

efekarakus:
Hi! We also encountered the same behavior as @bconway on 1.6.0: https://github.com/golang/mock/issues/571 which broke our tests

schmurfy:
I have the same issue, I reverted back to `1.5.0`, variadic arguments are useful when you don't care about the argument themselves but want to do something when called.

codyoss:
@schmurfy Please see my comments on #571

schmurfy:
@codyoss 
I really not sure I agree with this being by design but I will survive, using `Do(...interface{})` is a good way to clearly state in your test:
I don't give a sh*t about the arguments and they are irrelevant in this context

When you see it that way having variadic arguments does have a meaning, it is not just there because the author was lazy.

In one of my tests using it calling a method will trigger two parallel call to an external interface, I have a waitgroup and use `Do` to watch the external calls simply calling `wg.Done()` in them.

PR #562, opened 2021-05-28
Add ARM to support apple M1 chip

kishaningithub:


codyoss (approved):
Good idea, thanks for the contribution!

PR #564, opened 2021-06-04
fix ill-formatted message with fmt-verbs like %s

motemen:
If [fmt](https://pkg.go.dev/fmt)-verb strings such as `%s` is included in the error message, the output becomes `%!s(MISSING)` unintendedly.

This pull request fixes this problem.

### DIFF
--- a/gomock/callset.go
+++ b/gomock/callset.go
@@ -16,6 +16,7 @@ package gomock
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 )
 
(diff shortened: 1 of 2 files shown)


codyoss:
@motemen Could you please create an issue for the PR first for discussion. In that issue can you provide an example where this affects usage? Thanks.

motemen:
Thanks, filed a issue at https://github.com/golang/mock/issues/565 .

codyoss:
@motemen Thank you for your contribution 🎆

Issue #494, opened 2020-10-22
Unable to generate mocks in reflect mode


nezorflame:
**Actual behavior** _A clear and concise description of what the bug is._

My project uses Go modules, but due to the CI restrictions we also make use of the `vendor` folder.
So, our modules are synced by `go mod tidy && go mod vendor` command.

While using `mockgen` in a `reflect` mode, I'm unable to generate mocks. Can't pinpoint exactly when this became broken as this was working fine before, but it doesn't anymore.

The error message:

```
❯ go generate ./...

/Users/nezorflame/Go/src/github.com/nezorflame/mockgen-test - service - doc.go
prog.go:12:2: cannot find package "." in:
        /Users/nezorflame/Go/src/github.com/nezorflame/mockgen-test/vendor/github.com/golang/mock/mockgen/model
prog.go:12:2: cannot find package "." in:
        /Users/nezorflame/Go/src/github.com/nezorflame/mockgen-test/vendor/github.com/golang/mock/mockgen/model
prog.go:12:2: cannot find module providing package github.com/golang/mock/mockgen/model: working directory is not part of a module
prog.go:14:2: cannot find module providing package github.com/nezorflame/mockgen-test/internal/client: working directory is not part of a module
2020/10/22 19:41:04 Loading input failed: exit status 1
internal/service/doc.go:2: running "mockgen": exit status 1
```

Seems to be somewhat similar to the #423 but the proposed solution didn't work for me.
 
**Expected behavior** _A clear and concise description of what you expected to
happen._

Mocks should've been generated fine.

**To Reproduce** _Steps to reproduce the behavior_

I've published a [test project](https://github.com/nezorflame/mockgen-test) which can be used to reproduce the issue.

`mockgen` is called during the `go:generate` directive at [internal/service/doc.go](https://github.com/nezorflame/mockgen-test/blob/master/internal/service/doc.go):

```go
//go:generate echo $PWD - $GOPACKAGE - $GOFILE
//go:generate mockgen -package $GOPACKAGE -destination mocks_test.go github.com/nezorflame/mockgen-test/internal/client Producer
```

The interface to generate the mocks upon is pretty simple and is stored at [internal/client/contract.go](https://github.com/nezorflame/mockgen-test/blob/master/internal/client/contract.go):

```go

type Producer interface {
	Produce(ctx context.Context, objects []string) error
}
```

So, the steps are:

1. Download the project
2. (Optional) Generate the mocks without the `vendor`:
[...]

PR #567, opened 2021-06-04
add notes and error helper for vendor+reflect error

codyoss:
Fixes: #494

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,17 +30,6 @@ go install github.com/golang/mock/mockgen@v1.5.0
 If you use `mockgen` in your CI pipeline, it may be more appropriate to fixate
 on a specific mockgen version.
 
-## Documentation
-
-After installing, you can use `go doc` to get documentation:
-
-```bash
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -265,3 +254,23 @@ If the received value is `3`, then it will be printed as `03`.
...
+```
+
+If you come across this error while using reflect mode and vendoring
+dependencies there are three workarounds you can choose from:
+
+1. Use source mode.
+2. Include an empty import `import _ "github.com/golang/mock/mockgen/model"`.
+3. Add `--build_flags=--mod=mod` to your mockgen command.
+
+This error is do to changes in default behavior of the go command in more recent
```
bconway, line 274: Typo: s/do/due/

codyoss: Whoops, thank you for catching this! I will get it updated right away!

codyoss: https://github.com/golang/mock/pull/568

PR #569, opened 2021-06-08
fix parse array with the external const correctly

### LINKED ISSUE #561: Panic on array parsing if length is externally defined
**Actual behavior** 

generating code for the following interface makes the tool to panic. 

(Note: math constant used just for convenience. Can be any other package-defined constant, not a hardcoded value)
```go
type Fooer interface {
    Foo([math.MaxInt8]string)
}
```
Output: `panic: interface conversion: ast.Expr is *ast.SelectorExpr, not *ast.BasicLit`

**Expected behavior**

Generate a mocked struct with a method that takes a `[127]string{}` (or `[math.MaxInt8]string{}`) array

Explained in the "Actual behaviour" section

**Additional Information**

-   gomock mode (reflect or source): source
-   gomock version or git ref: v1.5.0
-   golang version: go version go1.15.8 darwin/amd64

~~**Triage Notes for the Maintainers**~~

sryoya:
Fixes #561 - Panic on array parsing if length is externally defined 

It also fixes the following minor problems in `TestParseArrayWithConstLength`, a test function related to this feature.
- It doesn't check all outputs to be asserted.
- It doesn't initialize some of the fields in `mockgen.fileParser`, even though they are initialized in the production code. (Actually, it caused an unnecessary test error by test data I added)

### DIFF
--- a/mockgen/parse.go
+++ b/mockgen/parse.go
@@ -416,7 +416,10 @@ func (p *fileParser) parseType(pkg string, typ ast.Expr) (model.Type, error) {
 			case (*ast.SelectorExpr):
 				// when the length is a const defined in an external package
 				usedPkg, err := importer.Default().Import(fmt.Sprintf("%s", val.X))
-				ev, err := types.Eval(token.NewFileSet(), usedPkg, token.NoPos, val.Sel.String())
+				if err != nil {
+					return nil, p.errorf(v.Len.Pos(), "unknown package in array length: %v", err)
+				}
+				ev, err := types.Eval(token.NewFileSet(), usedPkg, token.NoPos, val.Sel.Name)
 				if err != nil {
 					return nil, p.errorf(v.Len.Pos(), "unknown constant in array length: %v", err)
 				}
--- a/mockgen/parse.go
+++ b/mockgen/parse.go
@@ -22,8 +22,10 @@ import (
 	"fmt"
 	"go/ast"
 	"go/build"
+	"go/importer"
 	"go/parser"
 	"go/token"
+	"go/types"
 	"io/ioutil"
 	"log"
 	"path"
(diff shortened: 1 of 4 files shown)


mockgen/parse.go
```diff
@@ -409,8 +411,16 @@ func (p *fileParser) parseType(pkg string, typ ast.Expr) (model.Type, error) {
 			case (*ast.BasicLit):
 				value = val.Value
 			case (*ast.Ident):
-				// when the length is a const
+				// when the length is a const defined locally
 				value = val.Obj.Decl.(*ast.ValueSpec).Values[0].(*ast.BasicLit).Value
+			case (*ast.SelectorExpr):
+				// when the length is a const defined in an external package
+				usedPkg, err := importer.Default().Import(fmt.Sprintf("%s", val.X))
```
codyoss, line 418: nit: this error should be checked
nit: instead of `fmt.Sprintf("%s", val.X)` could you use val.X.Name

mockgen/parse.go
```diff
@@ -409,8 +411,16 @@ func (p *fileParser) parseType(pkg string, typ ast.Expr) (model.Type, error) {
...
+				ev, err := types.Eval(token.NewFileSet(), usedPkg, token.NoPos, val.Sel.String())
```
codyoss, line 419: nit: please use `val.Sel.Name`

codyoss (changes requested):
Thanks for the contribution, just a couple of minor nits.

mockgen/parse.go
sryoya: > nit: this error should be checked

Thanks for the catch! I missed this error.
https://github.com/golang/mock/pull/569/commits/d272a9382fc5b18efa25a2caa70b85b80435b839

> nit: instead of fmt.Sprintf("%s", val.X) could you use val.X.Name

The `X` field in the `SelectorExpr` type doesn't seem to have "Name" as its type is just the `Expr` interface.
https://github.com/golang/mock/pull/569/files

mockgen/parse.go
sryoya: Ok!

https://github.com/golang/mock/commit/d272a9382fc5b18efa25a2caa70b85b80435b839

sryoya:
Thanks for the review 🌵 I've either fixed or replied to your comments.

mockgen/parse.go
codyoss: Sorry I should have been more clear, you are right. To get at name you must cast it. val.X should always be an identifier in this case so `val.X.(*ast.Ident).Name` should work. But this is good to, less panic prone should ast representation change I suppose.

codyoss (approved):
LGTM

codyoss:
@sryoya Thank you for your contribution! 🎆

mockgen/parse.go
sryoya: Ah, I see. Your way looks fine too, but I feel using `fmt.Sprintf` is less panic prone as you say.
Thanks for the clarification 😄

PR #572, opened 2021-06-12
Move argument requirements for Do and DoAndReturn out of example code and into function commentary

bconway:
Regardless of the result of #558, this is not a new requirement (I believe), and the location of it in the example code was easy to miss.

The `Do` version excludes a reference to the output argument count, as the returns are ignored.


codyoss (approved):
Thanks for your contribution!

Issue #554, opened 2021-05-12
mockgen: record invocation arguments in generated code


rogpeppe:
It's not always obvious how `mockgen`-generated code has been created,
particularly when coming to it without prior knowledge of it. Good practice
would be to include a `go:generate` directive that can be used to regenerate
the code, but that's not always done.

Currently the comments on the generated code look something like this:

```
// Code generated by MockGen. DO NOT EDIT.
// Source: github.com/foo/bar (interfaces: Baz)

// Package mocks is a generated GoMock package.
```

I propose something along the lines of this:

```
// Code generated by MockGen. DO NOT EDIT.
// Source: github.com/foo/bar (interfaces: Baz)
//
// Generated by this command:
//    gomock github.com/foo/bar Baz

// Package mocks is a generated GoMock package.
```

codyoss:
Hey, thanks for the feature request, I like this idea! I think it would also be useful to record what version of mockgen generated the mock as well.

I would be happy to accept a PR for such a feature 😃

n1lesh:
@codyoss Would you be able to review the PR (#574)  I've submitted for this issue?

PR #574, opened 2021-06-12
Add invocation args & mockgen version in generated code

n1lesh:
Records invocation arguments and mockgen version (if exists) used to generate the mock
as per: #554


codyoss:
@n1lesh Would you mind adding a test showing what this would like like in generated files. Thanks.

codyoss (changes requested):
@n1lesh Would you mind adding a test showing what this would like like in generated files. Thanks.

codyoss:
I think this is still a good idea, but closing due to lack of response in case someelse wants to pick this up.

PR #583, opened 2021-08-28
Add go 1.17.x to the matrix build

kishaningithub:


kishaningithub:
When i ran this in my local with go 1.17 gofmt did not produce any error though. Any clues?

kishaningithub:
Checked this with the docker container as well. (Screenshot below)

Screenshot 2021-08-30 at 10 28 19 AM

codyoss:
Hey @kishaningithub Thanks for the PR. This is due to changes in how build constants are described in newer versions of Go. If you run `goimports -w .` from the repo root I think that should apply all the corrections needed. (Assuming you have 1.17 installed). Also would you mind removing 1.16 from the matrix. I think it is fine to test the oldest version we are compatible with and the latest.

codyoss (changes requested):
See comment above

codyoss:
Close this PR, I will add 1.17 when I have the chance.

PR #585, opened 2021-09-03
Updated README with correct information.

Saurabh-Thakre:
`go install` is not working with `path@version` and can only be used with `go get` syntax. Following is the error you get when you use `go install github.com/golang/mock/mockgen@v1.6.0`

> package github.com/golang/mock/mockgen@latest: can only use path@version syntax with 'go get'

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ GO111MODULE=on go get github.com/golang/mock/mockgen@v1.6.0
 ### Go 1.16+
 
 ```bash
-go install github.com/golang/mock/mockgen@v1.6.0
+go get github.com/golang/mock/mockgen@v1.6.0
 ```
 
 If you use `mockgen` in your CI pipeline, it may be more appropriate to fixate


Saurabh-Thakre:
@googlebot I signed it!

codyoss:
Hey @Saurabh-Thakre thanks for the PR. I think this all depends on the version of Go you are using. Which version do you currently have installed out of curiosity? I believe with Go 1.17 the suggestion is to use the `@version` syntax when installing binaries. This way it can work both in and out of the context of a module.

Saurabh-Thakre:
> Hey @Saurabh-Thakre thanks for the PR. I think this all depends on the version of Go you are using. Which version do you currently have installed out of curiosity? I believe with Go 1.17 the suggestion is to use the `@version` syntax when installing binaries. This way it can work both in and out of the context of a module.

Hey @codyoss, ohh it depends on the Go version, I thought this at first. by the way, I'm on `go version go1.15.8 linux/amd64`.  The following has got me confused somehow
### Go 1.16+

does it make sense if we modify it like this. Just a suggestion for a better understanding
### For  Go Version > 1.16

codyoss:
Does the provided command for your version of Go work for you? `GO111MODULE=on go get github.com/golang/mock/mockgen@v1.6.0`?

bradynpoulsen:
For what it's worth, in Go 1.17, it might make sense to use the new module-aware `go run` syntax with versioning:

```go
//go:generate github.com/golang/mock/mockgen@v1.6.0 -package ... -source ... -destination ...```

Saurabh-Thakre:
> Does the provided command for your version of Go work for you? `GO111MODULE=on go get github.com/golang/mock/mockgen@v1.6.0`?

Yeah it works as expected !

PR #586, opened 2021-09-07
AST duplication compare only by paths, not by alias. Fix it

TheZeroSlave:
HI.
AST removing duplicates works not correct for local aliasing.
If you have several imports ta package with different local names - it will bring some pain like "unknown package" because ast  remove all packages with another local naming. From the library source:
go/ast/filter.go:458
```if mode&FilterImportDuplicates != 0 {
		seen := make(map[string]bool)
		for _, filename := range filenames {
			f := pkg.Files[filename]
			for _, imp := range f.Imports {
				if path := imp.Path.Value; !seen[path] {
					// TODO: consider handling cases where:
					// - 2 imports exist with the same import path but
					//   have different local names (one should probably
					//   keep both of them)
					// - 2 imports exist but only one has a comment
					// - 2 imports exist and they both have (possibly
					//   different) comments
					imports = append(imports, imp)
					seen[path] = true
				}
			}
		}
```

So this PR tries to fix this by handling duplicates by skipping them.

### DIFF
--- a/mockgen/parse.go
+++ b/mockgen/parse.go
@@ -232,7 +232,7 @@ func (p *fileParser) parsePackage(path string) (*fileParser, error) {
 	}
 
 	for _, pkg := range pkgs {
-		file := ast.MergePackageFiles(pkg, ast.FilterFuncDuplicates|ast.FilterUnassociatedComments|ast.FilterImportDuplicates)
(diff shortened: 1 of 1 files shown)


TheZeroSlave:
@googlebot I signed it!

codyoss (changes requested):
Would you please  add a test and create an issue for this PR so it can be disscussed. In the issue it would be great to see a minimal example of what this is solving. Thanks!

codyoss:

Issue #174, opened 2018-04-15
Support variadic pointer parameter on caller.SetArg


rahmatismail:
gomock caller.SetArg recognizes variadic pointer parameter as a single param

## Code example
```// Some Sample Function
func (p *pkg) VariadicInterface(data ...interface{}) {
	for i := range data {
		data[i] = i
	}
}
func (p *pkg) VariadicInt(data ...*int) {
	for i := range data {
		*data[i] = i
	}
}

// Usage
var a, b, c, d int
pkgReal.VariadicInterface(&a, &b)
pkgReal.VariadicInt(&c, &d)
```

## Expect function on Test
```
mockPkg.EXPECT().VariadicInterface(gomock.Any(), gomock.Any()).SetArg(0, 5).SetArg(1, 5)
mockPkg.EXPECT().VariadicInt(gomock.Any(), gomock.Any()).SetArg(0, 5).SetArg(1, 5)
```

## Output
`SetArg(1, ...) called for a method with 1 args`

## Expected behaviour
Test passed with `a`, `b`, `c`, and `d` match given value

Please correct me if I use it wrong.

rahmatismail:
Sorry, I didn't notice this
`// TODO: This will break on variadic methods.`

PR #595, opened 2021-10-15
feat: allow setting args for variadic functions

bergotorino:
The gomock's Call interface allows for setting a value to a specified
argument, making it convinient to work with functions that return data
through pointer arguments. However, this is not possible for variadic
functions, the error is:

  SetArg(1, ...) called for a method with 1 args...

The root cause is that the variadic argument in a form of
...interface{} is threated as single value instead of being treated as
slice.

This commit changes this situation by making sure that the variadic
arguments that are a slice are treated as slices. As a consequence it is
now possible to set argument value for variadic functions.

See: https://github.com/golang/mock/issues/174

### DIFF
--- a/gomock/call.go
+++ b/gomock/call.go
@@ -225,15 +225,17 @@ func (c *Call) SetArg(n int, value interface{}) *Call {
 	c.t.Helper()
 
 	mt := c.methodType
-	// TODO: This will break on variadic methods.
-	// We will need to check those at invocation time.
-	if n < 0 || n >= mt.NumIn() {
+	if (n < 0 || n >= mt.NumIn()) && !mt.IsVariadic() {
 		c.t.Fatalf("SetArg(%d, ...) called for a method with %d args [%s]",
 			n, mt.NumIn(), c.origin)
 	}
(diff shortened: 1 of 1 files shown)


gomock/call.go
```diff
@@ -225,15 +225,17 @@ func (c *Call) SetArg(n int, value interface{}) *Call {
 	c.t.Helper()
 
 	mt := c.methodType
-	// TODO: This will break on variadic methods.
-	// We will need to check those at invocation time.
-	if n < 0 || n >= mt.NumIn() {
+	if (n < 0 || n >= mt.NumIn()) && !mt.IsVariadic() {
```
codyoss, line 228: Could you please add some tests for this. You can either do a traditional tests or a sample code generation like others in the `mockgen/internal/test` dir.

codyoss (changes requested):
Please add tests :)

bergotorino:
Sure, will do. Thanks for review and sorry for missing your comments on time

mvrahden:
@bergotorino are you still planning to do this?

codyoss:

PR #599, opened 2021-11-18
fix: avoid pkg name <> import pkg name conflicts

notbdu:
Avoid package name conflicts with an imported package name.


codyoss:
@notbdu  Please create an issue for this PR for discussion first. It would be good to see a minimal example of what this is addressing and see if it is needed. Thanks.

codyoss:

PR #603, opened 2021-12-10
fix: Support array parsing with length using binary expression and parenthesis

### LINKED ISSUE #575: Support array parsing with length using binary expression and parenthesis
*Requested feature* 
`Mockgen` should be able to parse an array using a binary expression or parenthesis in its length. They are not parsed correctly yet.
e.g. `[1 + 2]int`, `[(2)]int`

*Why the feature is needed*
As they are available expressions in the Go specifications, `Mockgen` should be able to handle them.

*(Optional) Proposed solution* A clear description of a proposed method for
More work needs to be done on the changes were made in #520 and #569

sryoya:
Fixes #575


codyoss (approved):
LGTM, sorry for the slow review!

Issue #606, opened 2021-12-23
Mockgen can't handle interface in current package other files


n0trace:
Mockgen can't handle interface in current package other files

Suppose I have two files:
1.  bar.go
```
package current

type Bar interface{}

```
2.  foo.go
```
package current

type Foo interface {
	Bar
	error
}
```
Running `mockgen -destination current_mock.go -package current -source=foo.go` gives
2021/12/23 21:37:55 Loading input failed: foo.go:4:2: unknown embedded interface Bar

```
    23  .  .  .  .  .  Type: *ast.InterfaceType {
    24  .  .  .  .  .  .  Interface: /home/work/go/src/github.com/golang/mock/foo.go:3:10
    25  .  .  .  .  .  .  Methods: *ast.FieldList {
    26  .  .  .  .  .  .  .  Opening: /home/work/go/src/github.com/golang/mock/foo.go:3:20
    27  .  .  .  .  .  .  .  List: []*ast.Field (len = 2) {
    28  .  .  .  .  .  .  .  .  0: *ast.Field {
    29  .  .  .  .  .  .  .  .  .  Type: *ast.Ident {
    30  .  .  .  .  .  .  .  .  .  .  NamePos: /home/work/go/src/github.com/golang/mock/foo.go:4:2
    31  .  .  .  .  .  .  .  .  .  .  Name: "Bar"
    32  .  .  .  .  .  .  .  .  .  }
    33  .  .  .  .  .  .  .  .  }
    34  .  .  .  .  .  .  .  .  1: *ast.Field {
    35  .  .  .  .  .  .  .  .  .  Type: *ast.Ident {
    36  .  .  .  .  .  .  .  .  .  .  NamePos: /home/work/go/src/github.com/golang/mock/foo.go:5:2
    37  .  .  .  .  .  .  .  .  .  .  Name: "error"
    38  .  .  .  .  .  .  .  .  .  }
    39  .  .  .  .  .  .  .  .  }
    40  .  .  .  .  .  .  .  }
    41  .  .  .  .  .  .  .  Closing: /home/work/go/src/github.com/golang/mock/foo.go:6:1
    42  .  .  .  .  .  .  }
    43  .  .  .  .  .  .  Incomplete: false
    44  .  .  .  .  .  }
```
result `Bar` like `error`

PR #607, opened 2021-12-23
feat: support interface in current package other files

n0trace:


n0trace:
#606

codyoss (approved):
LGTM, thanks for you contribution. This is possible today with aux-files flag but I would like to deprecate that flag in the future and this is a step towards that!

maxiloEmmmm:
need new release version, please

Issue #604, opened 2021-12-10
Only rewrite output files if changes are needed.


sodul:
**Actual behavior**
We have a fairly large project and running mockgen takes several minutes. During that time all of the generated mock files get deleted and rewritten. This causes 2 issues:
* if we run the mocks in parallel mockgen can try to read the files while they are being rewritten, causing mockgen to fail as it was trying to read the file. We worked around this by targeting directories in parallel but run mockgen sequentially within a directory.
* our IDEs see the files being modified and re-index the codebase over and over, making the IDE mostly unusable while mockgen is running.

**Expected behavior**
If the output file already exist and has the same content mockgen should not rewrite it and leave it untouched.

**To Reproduce** Steps to reproduce the behavior

Just run mockgen as usual.

**Additional Information**

- gomock 1.6.0
- golang 1.17.3

codyoss:
This seems like a reasonable feature request to me. I would be happy to accept a PR for such a change.

sodul:
@codyoss gentle ping. There are 2 potential PRs, small ones, that can address this issue.

FYI in our situation we have dozens, if not hundreds of generate mock files. So this change really help with the machines load since it reduce the re-indexing from IDEs, and reduces the background antivirus workload as well.

sodul:
@codyoss @sanposhiho gentle ping. I'm not sure who have merge permission rights on this repository.

sodul:
This was merged last month and should be in the final 1.7 release. @codyoss do you have a timeline for that?

amarjeetanandsingh:
Currently we are generating the mock and then compare if the generated content is already in the mock file. We are saving only the write efforts and it’s implications and not generation effort. I think there is a scope of improvement. I understand there will be challenges but can we consider below 2 approaches for optimisation?

- We can store the last modified UTC time of the actual file inside the generated file and use it to compare it on next mockgen trigger. 
- Another approach can be to store a fast hash of the actual file content inside the mock file and compare it on next mockgen trigger 

@codyoss @sodul 
If considerable, I’ll be happy to implement.

sodul:
@amarjeetanandsingh I'm not an official maintainer, I just contributed a PR here.

[...]

PR #618, opened 2022-02-08
add exists file data comparison before write

kozmod:
Add checking of destination file existence and comparison exists file data before rewriting the file: #604


sodul:
I have a very similar change in PR #626, which I wrote before I saw this was pending review. My version is slightly smaller, and I really do not mind which off the two versions get merged. @codyoss can pick either.

codyoss:
Closing in favor of #626

PR #626, opened 2022-02-22
Skip rewriting output files if unchanged

sodul:
When running mockgen with the output option this checks if the existing file content already exists and skips writing if there is nothing to change.

This will help reduce i/o when changing lots of files, but also reduce the re-indexing triggering in IDEs.

Solves #604

### DIFF
--- a/mockgen/mockgen.go
+++ b/mockgen/mockgen.go
@@ -165,7 +165,7 @@ func main() {
 		if err := os.MkdirAll(filepath.Dir(*destination), os.ModePerm); err != nil {
 			log.Fatalf("Unable to create directory: %v", err)
 		}
-		existing, err := os.ReadFile(*destination)
+		existing, err := ioutil.ReadFile(*destination)
 		if err != nil && !errors.Is(err, os.ErrNotExist) {
 			log.Fatalf("Failed reading pre-exiting destination file: %v", err)
 		}
--- a/mockgen/mockgen.go
+++ b/mockgen/mockgen.go
@@ -21,6 +21,7 @@ package main
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
 	"go/token"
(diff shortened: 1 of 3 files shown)


sodul:
Note that this duplicates an other pending PR #618. This one is slightly shorter but either one is fine by me @codyoss.

sodul:
@codyoss let me know if you need anything else to review here. Thank you!

sodul:
Gentle ping @codyoss This is a simple change that has a positive impact on IDEs and other tools that rescan on changed files.

mockgen/mockgen.go
```diff
@@ -171,7 +159,28 @@ func main() {
...
 		log.Fatalf("Failed generating mock: %v", err)
 	}
-	if _, err := dst.Write(g.Output()); err != nil {
+	output := g.Output()
+	dst := os.Stdout
+	if len(*destination) > 0 {
+		if err := os.MkdirAll(filepath.Dir(*destination), os.ModePerm); err != nil {
+			log.Fatalf("Unable to create directory: %v", err)
+		}
+		existing, err := os.ReadFile(*destination)
```
codyoss, line 168: This all seems reasonable to me, would you mind using an older API that is 1.15 compatible. After the next release I plan on dropping 1.15 support. But for now this would be nice to land in the next release.

codyoss (changes requested):
Just a small change

mockgen/mockgen.go
sodul: Done

mockgen/mockgen.go
```diff
@@ -171,7 +159,28 @@ func main() {
...
+	if len(*destination) > 0 {
+		if err := os.MkdirAll(filepath.Dir(*destination), os.ModePerm); err != nil {
+			log.Fatalf("Unable to create directory: %v", err)
+		}
+		existing, err := ioutil.ReadFile(*destination)
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
+			log.Fatalf("Failed reading pre-exiting destination file: %v", err)
+		}
+		if len(existing) == len(output) && bytes.Compare(existing, output) == 0 {
+			log.Println("Pre-existing file is already up to date.")
```
codyoss, line 173: nit: lets remove this log line. I think being silent here and just doing a noop is fine. I don't want to create noise in peoples CI where there is none today.

codyoss (changes requested):
Just one last nit and then I think this is good to go, thanks!

mockgen/mockgen.go
sodul: Done

codyoss (approved):
Thanks!

PR #632, opened 2022-03-25
Add -package_format to customize the generated package name

### LINKED ISSUE #631: Support customizing generated mock package name
Currently, mockgen allows specifying the generated mock package name with `-package` or a
generated package name with mock_ prefix will be used. However, when generating mock files
as package tests in scripts, this flag is very difficult to use.

In which case, the generated file shares the same package name as the package being mocked.
So we need to use tools (eg. grep) to extract the package name first, then pass it to mockgen.

This seems odd as mockgen already has the package name internally, and relies on external
tools for this are just a recipe for trouble.

I propose to introduce another flag to customize the generated mock package name, by default
it uses "mock_%s" which is the current behavior, and can be changed to other formats to allow
scripting usages.

dr-dime:
Previously `-package` must be explicitly defined or it will default
to `mock_${name}`. To generate a mock file that sits next to the
package files, eg. as part of the test, one must extract the package
name first. Which is not scripting-friendly.

After this change, we can use `-package_format` to customize the
generated package name. So that a script can run mockgen without
knowing the package name first.

Fixes #631

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,8 +77,10 @@ It supports the following flags:
   don't set this, the code is printed to standard output.
 
 - `-package`: The package to use for the resulting mock class
-  source code. If you don't set this, the package name is `mock_` concatenated
-  with the package of the input file.
+  source code. If you don't set this, the name is formatted with -package_format.
+
+- `-package_format`: The package name format when -package is not specified.
+  If you don't set this, `mock_%s` will be used.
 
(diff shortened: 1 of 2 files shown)


dr-dime:
@codyoss friendly ping.

dr-dime:
@codyoss friendly ping.

yuamera:
I also have the same needs, I hope to use it as soon as possible, thank you!

PR #633, opened 2022-03-25
Add archive mode

### LINKED ISSUE #424: Generate mocks from archives/*types.Type data
*Requested feature* 
Mockgen supports creating `github.com/golang/mock/mockgen/model.Interface` models from `reflect.Type` data using
https://github.com/golang/mock/blob/6d816de489c18a7e9a8fbd2aa5bb2dd75f2bbc86/mockgen/model/model.go#L295

But there is no analogue for creating `model.Interface`'s for mock generation from `go/types.Type` data.

*Why the feature is needed* 
The go compiler writes serialized type information for the public interface of each package, which can be read using [`golang.org/x/tools/go/gcexportdata`](https://pkg.go.dev/golang.org/x/tools/go/gcexportdata) to get a `types.Package`. 
This blog post gives a good indication of what the export data is good for: https://jayconrod.com/posts/112/export-data--the-secret-of-go-s-fast-builds.

We implemented a utility tool that takes an archive (that was generated by a previous build step from bazel), reads the type data, converts it to a gomock `model.Interface` and uses it for codegen. Using this instead of source/reflect mode presented significant efficiency and speed win for us over using gomock in reflect mode.

*(Optional) Proposed solution*
Although the internal tool works fine for us, there's some rough edges and some opportunity to share some of our benefits with the community. There's two things I have in mind for improvement:

- We maintain separate code for converting `types.Type` to `model.Interface`. This is pretty much analogous to whats implemented in `func InterfaceFromInterfaceType(it reflect.Type) (*model.Interface, error)` on https://github.com/golang/mock/blob/master/mockgen/model/model.go#L293-L470. `reflect.Type` and `types.Type` are similar (archives contain all the same typedata as the binary they later get compiled into), but unfortunately don't have an overarching interface. Would there be any interest in having support for converting types.Type` to `model.Interface` in golang/mock?
- golang/mock doesn't have any go api to take a `model.Package` and run codegen. Our current tool uses a bit of a workaround to do codegeneration: run gomock with `-exec_only` pointing to a stub binary, the stub binary reads the archive file, converts the `types.Package` into a `model.Package` and similar to a reflect binary encodes the `model.Package` with `encoding/gob` to a file for mockgen to read. In addition to its existing cli api for doing mockgen on a gob encoded `model.Package`, would it be possible to add a public go api to do codegen based on a `model.Package` object?

dr-dime:
This CL adds archive mode, which takes advantage of go/gcexportdata
and go/types packages to extract typing info from a package archive file.

Given the way go archive is designed, we don't need to worry about transitive
dependencies, as it already includes all transitive typing info. And, we don't
need to build or run anything, so even if cgo is involved, it will still work.

Fixes #424

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,20 @@ the version of mockgen used to generate your mocks.
 
 ## Running mockgen
 
-`mockgen` has two modes of operation: source and reflect.
+`mockgen` has three modes of operation: archive, source and reflect.
+
+### Archive mode
+
+Archive mode generates mock interfaces from a package archive
+file (.a). It is enabled by using the -archive flag, the import
+path is also needed as a non-flag argument. No other flags are
+required.
+
+Example:
+
+```bash
+mockgen -archive=pkg.a importpath
+```
 
 ### Source mode
 
@@ -70,6 +83,8 @@ mockgen . Conn,Driver
 The `mockgen` command is used to generate source code for a mock
 class given a Go source file containing interfaces to be mocked.
 It supports the following flags:
+
+- `-archive`: A package archive file containing interfaces to be mocked.
 
 - `-source`: A file containing interfaces to be mocked.
 
--- a/mockgen/archive.go
+++ b/mockgen/archive.go
@@ -1,0 +1,55 @@
+package main
+
+import (
+	"fmt"
+	"go/token"
+	"go/types"
+	"log"
+	"os"
+
+	"github.com/golang/mock/mockgen/model"
+
+	"golang.org/x/tools/go/gcexportdata"
+)
+
+func archiveMode(importpath, archive string) (*model.Package, error) {
+	f, err := os.Open(archive)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+	r, err := gcexportdata.NewReader(f)
+	if err != nil {
+		return nil, fmt.Errorf("read export data %q: %v", archive, err)
+	}
+
+	fset := token.NewFileSet()
+	imports := make(map[string]*types.Package)
+	tp, err := gcexportdata.Read(r, fset, imports, importpath)
+	if err != nil {
+		return nil, err
+	}
+
+	pkg := &model.Package{
+		Name:    tp.Name(),
+		PkgPath: tp.Path(),
+	}
+	for _, name := range tp.Scope().Names() {
+		m := tp.Scope().Lookup(name)
+		tn, ok := m.(*types.TypeName)
+		if !ok {
+			continue
+		}
+		ti, ok := tn.Type().Underlying().(*types.Interface)
+		if !ok {
+			continue
+		}
+		it, err := model.InterfaceFromGoTypesType(ti)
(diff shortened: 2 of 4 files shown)


linzhp:
Can you rebase this PR with main to resolve the conflicts and restart the tests?

README.md
```diff
@@ -33,7 +33,23 @@ the version of mockgen used to generate your mocks.
...
 ## Running mockgen
 
-`mockgen` has two modes of operation: source and reflect.
+`mockgen` has three modes of operation: archive, source and reflect.
+
+### Archive mode
+
+Archive mode generates mock interfaces from a package archive
+file (.a). It is enabled by using the -archive flag, the import
+path is also needed as a non-flag argument. No other flags are
```
linzhp, line 42: Can archive mode optionally take a list of interfaces, just like reflect mode? Mocking extra interfaces may introduce extra dependencies to the mock code.

mockgen/archive.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"fmt"
+	"go/token"
+	"go/types"
+	"log"
+	"os"
+
+	"github.com/golang/mock/mockgen/model"
+
```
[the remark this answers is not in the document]
dr-dime: I believe this follows google's Golang style guide for grouping imports, i.e. std, local then third-party.

README.md
dr-dime: Could be, but it also means this mode will derive from the source mode. We will need to manually match the specified interfaces to the resolved types. IMHO, the generated code will not introduce dependencies that are not already transitively depended on (in theory), so it should be just fine.

dr-dime:
> Can you rebase this PR with main to resolve the conflicts and restart the tests?

Done. PTAL.

linzhp (approved):
LGTM. @codyoss Can you merge?

README.md
linzhp: Saying we have:

```Go
package foo

type SFoo struct {
}

type IBar Interface {
  Method1() string
}

type IBaz interface {
  Method2() SFoo
}
```

When we only mock `IBar`, then the mock won't depend on package `foo`, so it can be compiled into a different package `foomock` and imported by the tests in package `foo`. However, when we mock both `IBar` and `IBaz`, if the mock is still compiled into `foomock`, then the mock would have to import package `foo` because a method in `IBaz` returns `foo.SFoo`. Now the tests in `foo` cannot import `foomock`, because of circular dependencies.

dr-dime: Well, in this case, I'll just put the generated mocks in foo, say mock_foo_test.go. 

I'm never a big fan of exporting mocks for unit tests in packages. The obvious reason is that you are actually testing against an imaginary counterpart, a passing test doesn't reflect the actual usage of the package being mocked.

If you need to provide something for external packages to test against, create a fake sharing the same logic.

This also follows  https://github.com/golang/go/wiki/CodeReviewComments#interfaces.

Issue #685, opened 2022-12-09
Proposal: Provide an option where every invocation of `EXPECT()` of a mock should override any previous invocations.


styeung:
Hi team,

The proposal here is to allow for an option to bring GoMock closer in line with other mocking libraries that exist for other languages. In particular, the GoMock behavior of ordering mock expectations in a FIFO manner makes it difficult to bundle default mock expectations into a helper or testify Suite's BeforeTest block.

## The problem
For example, the following set up won't work:

```go
type ExampleTestSuite struct {
    suite.Suite

	repoAMock *MockRepositoryA
	repoBMock *MockRepositoryB
	repoCMock *MockRepositoryC

	sut *Controller
}

func TestExampleTestSuite(t *testing.T) {
    suite.Run(t, new(ExampleTestSuite))
}

func (s *ExampleTestSuite) SetupTest() {
	ctrl := gomock.NewController(s.T())
	s.repoAMock = NewMockRepositoryA(ctrl)
	s.repoBMock = NewMockRepositoryB(ctrl)
	s.repoCMock = NewMockRepositoryC(ctrl)

	s.sut = NewController(s.repoAMock, s.repoBMock, s.repoCMock) // subject under test

	// set up mock defaults
	s.repoAMock.EXPECT().Get(gomock.Any()).Return("foo", nil)
	s.repoBMock.EXPECT().Get(gomock.Any()).Return("bar", nil)
	s.repoCMock.EXPECT().Get(gomock.Any()).Return("baz", nil)
}


func (s * ExampleTestSuite) TestGather_HappyPath() {
	res, rr := s.sut.Gather()
	s.Require().Equal("foo, bar, baz", res)
	s.Require().NoError(err)
}

func (s *ExampleTestSuite) TestGather_RepoAFails() {
	s.repoAMock.EXPECT().Get(gomock.Any()).Return("", assert.AnError)

	_, err := s.sut.Gather()
	s.Require().Error(err)
}

func (s *ExampleTestSuite) TestGather_RepoBFails() {
	s.repoBMock.EXPECT().Get(gomock.Any()).Return("", assert.AnError)

	_, err := s.sut.Gather()
	s.Require().Error(err)
}

func (s *ExampleTestSuite) TestGather_RepoCFails() {
	s.repoCMock.EXPECT().Get(gomock.Any()).Return("", assert.AnError)

	_, err := s.sut.Gather()
	s.Require().Error(err)
}
```

Instead, one must do something like this:

```go
[...]

PR #686, opened 2022-12-09
Implement OverridableController to allow for mock overrides

styeung:
This is an initial implementation of the proposal outlined in this issue, [Proposal: Provide an option where every invocation of EXPECT() of a mock should override any previous invocations. #685](https://github.com/golang/mock/issues/685)


shivangawasthi:
@codyoss is it possible to get this merged. Will be helpful.

musafir-V:
@codyoss can we get this merged? will be helpful.

PR #702, opened 2023-04-30
record invocation arguments in generated code

### LINKED ISSUE #554: mockgen: record invocation arguments in generated code

kucherenkovova:
resolves #554


kucherenkovova:
@codyoss please review

